PR #2, opened 2017-07-05
repo scanner works 

AnthonySteele:
I think


NuKeeper/RepositoryInspection/RepositoryScanner.cs
```diff
@@ -0,0 +1,59 @@
...
+            foreach (var subdir in subdirs)
+            {
+                var subdirPackages = FindNugetPackagesInDirRecursive(subdir);
+                current.AddRange(subdirPackages);
+            }
+
+            return current;
+        }
+
+        private List<NugetPackage> FindNugetPackagesInDir(string dir)
```
cohen990: ScanForNugetPackages(string currentDir) ?

PR #26, opened 2017-07-07
Scanning packages config

skolima:
This fixes scanning old style packages.config and improves performance by only contacting nuget.org once per package Id.

Unfortunately, this is not yet enough to work for old style projects (at least for me), as `dotnet add package` still fails for me with

```
error: Value cannot be null.
error: Parameter name: projectUniqueName
```


AnthonySteele:
It might be that for old style projects, we should not use the `dotnet` command at all, but the nuget command line instead, e.g. `nuget update {project} -Id {packageId} -Version {packageVersion} -NonInteractive`

PR #30, opened 2017-07-08
Just ask directly for files we're interested in.

skolima:
No need to enumerate the whole directory.

I realised I'll need to re-do this on once #28 is merged.


AnthonySteele:
Fair enough, but we should merge #28 first

cohen990:
like both of you have said, this needs to be re-done :P

skolima:
Re-done on latest master

AnthonySteele:
👍

PR #31, opened 2017-07-08
Updater for packages.config files

skolima:
This follows on #30 change

Uses NuGet.exe installed as a package, however, there's a problem: behaviour of this tool is not the same as updating through Visual Studio UI. Main difference seems to be that install/uninstall PowerShell scripts aren't run.

Examples:
~https://www.nuget.org/packages/System.IO.Abstractions/ would add app.config assembly mappings, but command line doesn't do this~ for some reason this doesn't happen any more from UI either
https://www.nuget.org/packages/StyleCop.Analyzers/ would update analyzers in .csproj, but command line doesn't do this

### DIFF
--- a/NuKeeper.Tests/RepositoryInspection/PackagesFileReaderTests.cs
+++ b/NuKeeper.Tests/RepositoryInspection/PackagesFileReaderTests.cs
@@ -58,6 +58,7 @@ namespace NuKeeper.Tests.RepositoryInspection
             Assert.That(package, Is.Not.Null);
             Assert.That(package.Id, Is.EqualTo("foo"));
             Assert.That(package.Version, Is.EqualTo(new NuGetVersion("1.2.3.4")));
+            Assert.That(package.PackageType, Is.EqualTo(PackageType.PackagesConfig));
         }
 
         [Test]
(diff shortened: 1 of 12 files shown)


NuKeeper/Nuget/Process/PackagesConfigUpdater.cs
```diff
@@ -0,0 +1,46 @@
...
+            var updateCommand = $"cd {dirName}"
+                + $" & {nuget} restore packages.config"
+                + $" & {nuget} update packages.config -Id {update.PackageId} -Version {update.NewVersion}";
+            Console.WriteLine(updateCommand);
+            await RunExternalCommand(updateCommand);
+        }
+
+        private async Task RunExternalCommand(string command)
+        {
+            var result = await _externalProcess.Run($"{command}");
```
AnthonySteele: `$"{command}"` is exactly equivalent to `command`.

AnthonySteele:
Very good, but we should also merge #28 first

cohen990:
Needs re-doing following merge of #28

skolima:
Re-done on current master

PR #36, opened 2017-07-09
Also build for Net462

AnthonySteele:
Also build a windows executable using .Net 4.6.2
Would be easier if there was just one EasyConfig.Net version!

### DIFF
--- a/NuKeeper/NuKeeper.csproj
+++ b/NuKeeper/NuKeeper.csproj
@@ -1,13 +1,18 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp1.1</TargetFramework>
+    <TargetFrameworks>netcoreapp1.1;Net462</TargetFrameworks>
     <RootNamespace>NuKeeper</RootNamespace>
     <AssemblyName>NuKeeper</AssemblyName>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="EasyConfig.Net.Core" Version="1.0.6" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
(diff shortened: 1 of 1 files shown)


NuKeeper/NuKeeper.csproj
```diff
@@ -1,13 +1,18 @@
...
   <ItemGroup>
-    <PackageReference Include="EasyConfig.Net.Core" Version="1.0.6" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Include="NuGet.Protocol.Core.v3" Version="4.0.0" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.1' ">    
+	<PackageReference Include="EasyConfig.Net.Core" Version="1.0.6" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'Net462' ">    
+	<PackageReference Include="EasyConfig.Net" Version="1.1.22" />
```
cohen990: concerned about this - it'll be better if we can use EasyConfig.Net.Core instead of EasyConfig.Net since they're different packages and have already begun to diverge

AnthonySteele: agreed that it's messy.
I think that what has to happen is that the easyconfig lib has to also have `<TargetFrameworks>netstandard1.6;Net452</TargetFrameworks>` or similar in the csproj.
i.e. make the nuget package contain build artefacts for both platforms.

cohen990: Have created an issue in EasyConfig https://github.com/cohen990/EasyConfig.Net.Core/issues/2

AnthonySteele: Close this PR for now?

cohen990: I don't entirely know when I'll be able to get to that fix in EasyConfig...

PR #39, opened 2017-07-11
Clean-up previous runs at startup

AnthonySteele:
Clean-up previous runs at startup
This works, to an extent. It deletes old files that aren't locked.


cohen990:
We really need to get some CI

Issue #21, opened 2017-07-07
Need consistency between command line arguments


cohen990:
Currently, Repository Mode accepts a repository URI and extracts the owner, repo name and github api endpoint from that.

Organisation mode however accepts organisation name and github api endpoint.

We want to standardise these

AnthonySteele:
Also want to move away from positional args except possibly for the first mode param, and use e.g. `-token foo -repo bar`.  
Also look at  a lib for parsing these.
And values for env vars and./or config file.

cohen990:
We've moved away from positional arguments now - see #25

AnthonySteele:
Close or?

PR #40, opened 2017-07-11
RepositoryMode can use the GithubApiEndpoint from commandline

AnthonySteele:
RepositoryMode can use the GithubApiEndpoint from commandline
if it is supplied

### DIFF
--- a/NuKeeper/Configuration/CommandLineParser.cs
+++ b/NuKeeper/Configuration/CommandLineParser.cs
@@ -35,8 +35,18 @@ namespace NuKeeper.Configuration
             var githubToken = settings.GithubToken;
             var githubRepoUri = settings.GithubRepositoryUri;
 
+            Uri githubApiBase;
+            if (settings.GithubApiEndpoint != null)
+            {
+                githubApiBase = settings.GithubApiEndpoint;
+            }
+            else
+            {
+                var generatedApiBase = "https://api." + githubRepoUri.Host;
+                githubApiBase = new Uri(generatedApiBase);
+            }
+
             // general pattern is https://github.com/owner/reponame.git
-            var githubHost = "https://api." + githubRepoUri.Host;
             var path = githubRepoUri.AbsolutePath;
             var pathParts = path.Split('/')
                 .Where(s => !string.IsNullOrWhiteSpace(s))
@@ -49,7 +59,7 @@ namespace NuKeeper.Configuration
             {
                 GithubUri = githubRepoUri,
                 GithubToken = githubToken,
-                GithubApiBase = new Uri(githubHost),
+                GithubApiBase = githubApiBase,
                 RepositoryName = repoName,
                 RepositoryOwner = repoOwner
             };


NuKeeper/Configuration/CommandLineParser.cs
```diff
@@ -35,8 +35,18 @@ private static RepositoryModeSettings ReadSettingsForRepositoryMode(CommandLineA
             var githubToken = settings.GithubToken;
             var githubRepoUri = settings.GithubRepositoryUri;
 
+            Uri githubApiBase;
+            if (settings.GithubApiEndpoint != null)
+            {
+                githubApiBase = settings.GithubApiEndpoint;
```
cohen990: do we want to take the repository name as input? or do we still want to parse it out of a uri - it leaves them having to post an argument like this 

`dotnet run mode=repository github_token=<GitToken> github_repository_uri=https://github.com/NuKeeperDotNet/NuKeeper github_api_endpoint=https://api.github.com`

rather than 

`C:\Code\NuKeeper\NuKeeper>dotnet run mode=organisation github_token=<GitToken> github_api_endpoint=https://api.github.com/ github_organisation_name=NuKeeperDotNet github_repository_name=NuKeeper `

wdyt?

AnthonySteele: In your e.g. this isn't needed for github.com as code to generate the api url was written for github.com. It doesn't work on an internal github with a different api url structure.

I agree that we want to standardise args (issue #21) and use something like the organisation mode version. How far away is this? If it's near than I will withdraw this. Otherwise, it allows me to test other things today in the meantime.

skolima: Isn't this what the default argument values are for? Have `github_api_endpoint=https://api.github.com` as default and you can override it if you need to?

cohen990: we aren't taking [this line](https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper/Configuration/CommandLineParser.cs#L53) into account - we need the raw github endpoint in order to clone it - unless we get smarter and query github using the org and reponame.

I think for now we have to specify both the api endpoint and the repo uri

AnthonySteele: Yeah, we would have to tie this mode to github integration as well, and use the github api to find the url to pull from & push to.

NuKeeper/Configuration/CommandLineParser.cs
```diff
@@ -35,8 +35,18 @@ private static RepositoryModeSettings ReadSettingsForRepositoryMode(CommandLineA
             var githubToken = settings.GithubToken;
             var githubRepoUri = settings.GithubRepositoryUri;
 
+            Uri githubApiBase;
+            if (settings.GithubApiEndpoint != null)
```
cohen990: will never be null if we merge #44

PR #41, opened 2017-07-11
easyconfig configuration error handling to handle missing arguments

cohen990:
The error handling will end up with a throw (and a dotnet core crash)

If that's acceptable to us, then easyconfig already has code in place for informative error handling (even if the elegance of it is questionable)

example

### DIFF
--- a/NuKeeper/Configuration/CommandLineArguments.cs
+++ b/NuKeeper/Configuration/CommandLineArguments.cs
@@ -5,7 +5,7 @@ namespace NuKeeper.Configuration
 {
     public class CommandLineArguments
     {
-        [CommandLine("mode")]
+        [CommandLine("mode"), Required]
         public string Mode;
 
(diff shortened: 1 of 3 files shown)


AnthonySteele:
Perhas that should have a `try...catch(ConfigurationException) {}` in the program to output the exception message and exit(1) ?  Other than that, I am ok with it.

skolima:
As long as those messages are informative, I'm ok with this. Just no `IndexOutOfBounds` or other built-in exceptions.

cohen990:
@skolima the attached image details the type of error I'm talking about. There may still be edge cases that I've not considered (I'm sure there are) but they need to be fixed in EasyConfig, not compensated for in NuKeeper

AnthonySteele:
👍 Looks ok to me. Is it ready to merge?

cohen990:
It is now @AnthonySteele

PR #42, opened 2017-07-11
Automatic update of EasyConfig.Net.Core to 1.0.7

cohen990:
NuKeeper has generated an update of `EasyConfig.Net.Core` from `1.0.6` to `1.0.7`
One project update:
Updated `NuKeeper\NuKeeper.csproj` to `EasyConfig.Net.Core` `1.0.7` from `1.0.6`
This is an automated update. Merge only if it passes tests

**NuKeeper**: https://github.com/NuKeeperDotNet/NuKeeper

### DIFF
--- a/NuKeeper/NuKeeper.csproj
+++ b/NuKeeper/NuKeeper.csproj
@@ -6,7 +6,7 @@
     <AssemblyName>NuKeeper</AssemblyName>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


cohen990:
Closing since a new version is incoming

AnthonySteele:
We should also look at "from 1.0.6 to 1.0.7" vs "to EasyConfig.Net.Core 1.0.7 from 1.0.6"
My thinking is that the new version comes first, e.g. "too 1.0.7 from 1.0.6"

PR #53, opened 2017-07-13
Minor prefactor

AnthonySteele:
smaller refactor before adding a new class  …
mostly:
- extract methods in RepositoryUpdater
- code that writes out goes into EngineReport and is a bit better


cohen990:
Its out of date. Apart from that I'm happy

AnthonySteele:
Done. Internet was down when I started this, hence the stale master.

PR #56, opened 2017-07-15
Package update set

AnthonySteele:
Introduce the class `PackageUpdateSet` which encapsulates the data for 1 package that is going to be updated to a new version, in 1 or more existing `PackageInProject`

Use it instead of a list of `PackageUpdate` for the same new package in different projects
Deleted class `PackageUpdate`.


NuKeeper.Tests/RepositoryInspection/PackageUpdateSetTests.cs
```diff
@@ -0,0 +1,122 @@
...
+using NuKeeper.RepositoryInspection;
+using NUnit.Framework;
+
+namespace NuKeeper.Tests.RepositoryInspection
+{
+    [TestFixture]
+    public class PackageUpdateSetTests
+    {
+        [Test]
+        public void NullNewPackageId_IsNnotAllowed()
```
cohen990: "NullNewPackageId_IsNnotAllowed" -> "NullNewPackageId_IsNotAllowed"

NuKeeper.Tests/RepositoryInspection/PackageUpdateSetTests.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        [Test]
+        public void CannotHaveUpdateForDifferentPackageToNewVersion()
+        {
+            var newPackage = NewPackageId();
+
+            var currentPackages = new List<PackageInProject>
+            {
+                new PackageInProject("bar", "1.0.0", PathToProjectOne())
```
cohen990: the code for this test doesn't explicitly declare that "bar" is different to "foo" - have to f12 around for that. Is there a way of making the code of this test more declarative?

e.g. `NewPackageId()` -> `NewPackageWithId("foo")`

NuKeeper.Tests/RepositoryInspection/PackagesFileReaderTests.cs
```diff
@@ -69,8 +70,14 @@ public void TwoPackagesShouldBeRead()
...
             Assert.That(packages, Is.Not.Null);
             Assert.That(packages.Count, Is.EqualTo(2));
+
+            PackageAssert.IsPopulated(packages[0]);
             Assert.That(packages[0].Id, Is.EqualTo("foo"));
+            Assert.That(packages[0].Version, Is.EqualTo(new NuGetVersion("1.2.3.4")));
+
+            PackageAssert.IsPopulated(packages[1]);
             Assert.That(packages[1].Id, Is.EqualTo("bar"));
+            Assert.That(packages[1].Version, Is.EqualTo(new NuGetVersion("2.3.4.5")));
```
cohen990: This is a lot of asserts for a single test. 16 of them if we unpack `PackageAssert.IsPopulated`

cohen990: There's a good discussion here https://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test

In particular:
> Tests should fail for one reason only, but that doesn't always mean that there should be only one Assert statement. IMHO it is more important to hold to the "Arrange, Act, Assert" pattern.


This test does seem fairly solid in that it's not going to fail for many different reasons. Maybe the test is too big? Maybe the function behind needs breaking down?

16 asserts makes me nervous - has to be a code smell...

NuKeeper.Tests/RepositoryInspection/ProjectFileReaderTests.cs
```diff
@@ -136,7 +142,7 @@ public void ResultIsReiterable()
 
             foreach (var package in packages)
             {
-                Assert.That(package, Is.Not.Null);
+                PackageAssert.IsPopulated(package);
             }
 
             Assert.That(packages.Select(p=>p.Path), Is.All.EqualTo(path));
```
cohen990: While I agree that this test will test whether or not the result is iterable, it has extra detail in the tests that could confuse the intent for somebody reading the test

NuKeeper/NuGet/Process/NuGetUpdater.cs
```diff
@@ -14,10 +15,10 @@ public NuGetUpdater(IExternalProcess externalProcess = null)
             _externalProcess = externalProcess ?? new ExternalProcess();
         }
 
-        public async Task UpdatePackage(PackageUpdate update)
+        public async Task UpdatePackage(PackageIdentity newPackage, PackageInProject currentPackage)
```
cohen990: I don't like that newPackage and currentPackage are different types - seems misleading.

Is there a reasonable renaming that can be done to alleviate that?

`newPackage` -> `newPackageIdentity`?

NuKeeper/Engine/RepositoryUpdater.cs
```diff
@@ -73,57 +74,54 @@ public RepositoryUpdater(IPackageUpdatesLookup packageLookup, IGithub github, Re
...
+        private async Task UpdatePackageInProjects(PackageUpdateSet updateSet)
         {
-            var oldVersions = updates
-                .Select(u => u.OldVersion.ToString())
+            var oldVersions = updateSet.CurrentPackages
+                .Select(u => u.Version.ToString())
                 .Distinct();
 
             var oldVersionsString = string.Join(",", oldVersions);
+            var packageId = updateSet.NewPackage.Id;
```
cohen990: use `updateSet.PackageId`

NuKeeper.Tests/RepositoryInspection/PackagesFileReaderTests.cs
AnthonySteele: Fair comment, have updated to split up some of those tests into smaller parts

NuKeeper/NuGet/Process/NuGetUpdater.cs
AnthonySteele: So a `PackageIdentity` is a type built into the nuget lib, consists of a package name and version. 
I'm thinking of changing this around even more...

NuKeeper/Engine/RepositoryUpdater.cs
AnthonySteele: Done, and moved to EngineReport

PR #63, opened 2017-07-18
Update priority

AnthonySteele:
Extracted the class `PackageUpdateSelection`  out of `RepositoryUpdater`
it decides which of the possible package update sets should be applied. 
I.e. it applies the max pr limit form the settings.
Also, it has a (very basic) sort order to decide which update is prefered.
At present it's just take those with the most existing versions, i.e. "Consolidate" first.
`PackageUpdateSet` has the method `CountCurrentVersions` for this.

Naming suggestions are welcome.
Will likely sort on other things as well later (e.g. number of package usages).


NuKeeper/Engine/PackageUpdateSelection.cs
```diff
@@ -0,0 +1,30 @@
...
+using NuKeeper.Configuration;
+using NuKeeper.RepositoryInspection;
+
+namespace NuKeeper.Engine
+{
+    public class PackageUpdateSelection : IPackageUpdateSelection
+    {
+        private readonly RepositoryModeSettings _settings;
+
+        public PackageUpdateSelection(RepositoryModeSettings settings)
```
cohen990: can we cut this down to just pass in the max pull requests?

AnthonySteele: We can do that. Another look at settings might be a later goal.

cohen990: Yeah I'm kinda unhappy with the way it's all set up now - so that's another reason to not pass it round everywhere :P

NuKeeper/Configuration/Settings.cs
```diff
@@ -31,6 +31,8 @@ public Settings(RepositoryModeSettings repository, OrganisationModeSettings orga
         public string GithubToken => Repository?.GithubToken ?? Organisation?.GithubToken;
         public Uri GithubApiBase => Repository?.GithubApiBase ?? Organisation?.GithubApiBase;
 
+        public int MaxPullRequestsPerRepository => Repository?.MaxPullRequestsPerRepository ?? Organisation?.MaxPullRequestsPerRepository ?? 0;
```
cohen990: haha - yeah we have to revisit this soon

PR #73, opened 2017-07-24
Add LibGit2Sharp support

### LINKED ISSUE #10: Use a git lib not command line
Maybe https://github.com/libgit2/libgit2sharp

slang25:
Fixes #10

### DIFF
--- a/NuKeeper/Git/LibGit2SharpDriver.cs
+++ b/NuKeeper/Git/LibGit2SharpDriver.cs
@@ -1,0 +1,80 @@
+using System;
+using System.Threading.Tasks;
+using LibGit2Sharp;
+
+namespace NuKeeper.Git
+{
+    public class LibGit2SharpDriver : IGitDriver
+    {
+        private readonly string _repoStoragePath;
+        private readonly string _githubUser;
+        private readonly string _githubToken;
+
+        public LibGit2SharpDriver(string repoStoragePath, string githubUser, string githubToken)
+        {
+            _repoStoragePath = repoStoragePath;
+            _githubUser = githubUser;
+            _githubToken = githubToken;
+        }
+        public Task Clone(Uri pullEndpoint)
+        {
+            Repository.Clone(pullEndpoint.ToString(), _repoStoragePath);
+            return Task.CompletedTask;
+        }
+
+        public Task Checkout(string branchName)
+        {
+            using (var repo = new Repository(_repoStoragePath))
+            {
+                Commands.Checkout(repo, repo.Branches[branchName]);
+            }
+            return Task.CompletedTask;
+        }
+
+        public Task CheckoutNewBranch(string branchName)
+        {
+            using (var repo = new Repository(_repoStoragePath))
+            {
+                var branch = repo.CreateBranch(branchName);
+                Commands.Checkout(repo, branch);
+            }
+            return Task.CompletedTask;
+        }
+
+        public Task Commit(string message)
+        {
+            using (var repo = new Repository(_repoStoragePath))
+            {
+                var sig = repo.Config.BuildSignature(DateTimeOffset.Now);
+                Commands.Stage(repo, "*");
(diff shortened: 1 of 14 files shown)


NuKeeper/TempFiles.cs
```diff
@@ -12,7 +13,8 @@ private static string NuKeeperTempFilesPath()
 
         public static void DeleteExistingTempDirs()
         {
-            var dirs = Directory.EnumerateDirectories(NuKeeperTempFilesPath());
+            var dirInfo = new DirectoryInfo(NuKeeperTempFilesPath());
+            var dirs = dirInfo.Exists ? dirInfo.EnumerateDirectories() : Enumerable.Empty<DirectoryInfo>();
```
slang25: I made this change cause it doesn't work on first use otherwise

slang25:
I think I may not have needed to add the username command line argument, as the `CredentialsProvider` has an argument called `usernameFromUrl`, however I'm a little confused by how this should work.

I point it at a repo that belongs to an org, I was expecting this to create a fork, we push to that fork, then the PR is from that fork, in which case `usernameFromUrl` would work, however I don't think this is the case, I think this always pushes the code to a branch on the org repo, and creates a PR from there, this means we'll need to pass the username through and have to have some permissions on the org repo.

README.md
```diff
@@ -43,6 +43,7 @@ C:\Code\NuKeeper\NuKeeper>dotnet run mode=organisation github_token=<GitToken> g
 | Name                             | Alias | Default value          | Required |
 |----------------------------------|-------|------------------------|-----------------|
 | mode                             | m     |                        | Yes             |
+| github_user                      | u     |                        | Yes             |
```
cohen990: Need to also add an explanation of parameter below and add the parameter to the examples above

AnthonySteele:
Why do we need a github user?

slang25:
We don't, my bad.

slang25:
@AnthonySteele Updated, I was rushing though 🤞

NuKeeper/Git/LibGit2SharpDriver.cs
```diff
@@ -0,0 +1,80 @@
...
+        public LibGit2SharpDriver(string repoStoragePath, string githubUser, string githubToken)
+        {
+            _repoStoragePath = repoStoragePath;
+            _githubUser = githubUser;
+            _githubToken = githubToken;
+        }
+        public Task Clone(Uri pullEndpoint)
+        {
+            Repository.Clone(pullEndpoint.ToString(), _repoStoragePath);
+            return Task.CompletedTask;
```
AnthonySteele: Is `LibGit2Sharp` likely to become async at any time soon?
Basically if not, and we discard the other git driver we should then make these methods sync.

slang25: Yeah, not anytime soon is my understanding https://github.com/libgit2/libgit2sharp/issues/1440

slang25: Shall I do that in this PR?

AnthonySteele: Can cleanup in a following PR.

NuKeeper/Program.cs
```diff
@@ -27,7 +28,12 @@ public static int Main(string[] args)
             var repositoryDiscovery = new GithubRepositoryDiscovery(github, settings);
             var updateSelection = new PackageUpdateSelection(settings.MaxPullRequestsPerRepository);
 
-            RunAll(repositoryDiscovery, lookups, updateSelection, github)
+            // get some storage space
+            var tempDir = TempFiles.MakeUniqueTemporaryPath();
+            var githubUser = github.GetCurrentUser().GetAwaiter().GetResult();
```
AnthonySteele: I would prefer to push these lines into a single async method so that there's only one point of resynchronisation with `.GetAwaiter().GetResult()`

PR #82, opened 2017-07-26
Changes to config

cohen990:


README.md
```diff
@@ -27,35 +27,54 @@ Will PR a single repository. Point it at the https url of a github repository, l
...
+C:\Code\NuKeeper\NuKeeper>dotnet run mode=organisation t=<GitToken> github_api_endpoint=https://api.github.com/ github_organisation_name=<OrgName>
 ```
+### Environment Variables
+
+| Name                             | Required          | Overridable via CLI? |
+|----------------------------------|-------------------|----------------------|
+| NuKeeper_github_token            | If not overridden | Yes (`t`)            |
+
+ * *NuKeeper_github_token* You will need to [create a github personal access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/) to authorise access to your github server in order to raise PRs. Be sure to check the "repo" scope when creating the token. If you have just created this environment variabled, remember to restart your terminal and IDE before proceeding.
+
```
AnthonySteele: Typo "variabled"

AnthonySteele:
Agree with the idea - some of these values can live / have a default in the config file.
The Api token can be in an env var (with override, as we work with more than one github server).

cohen990:
The command line override for the token is mostly for those people who object to adding environment variables for every single program they run/install

NuKeeper/Configuration/CommandLineArguments.cs
```diff
@@ -8,7 +9,8 @@ public class CommandLineArguments
         [CommandLine("mode", "m"), Required]
         public string Mode;
 
-        [CommandLine("github_token", "t"), Required, SensitiveInformation]
+        [Environment("NuKeeper_github_token"), Required, SensitiveInformation]
+        [OverriddenBy(ConfigurationSources.CommandLine, "t")]
```
AnthonySteele: does this mean that you can *only* use the short name on the command line? I would prefer to keep both.

cohen990: I thought it would be very verbose to include a full name AND an alias for the override (not to mention, certain sources are better without an alias e.g. environment or config file)

cohen990: I'm open to suggestions though

AnthonySteele: Agreed that it's only the command line where "short names" is even a common thing. I'm wondering if we want them there.

cohen990: maybe if we're overriding, we don't want it to be necessarily easy - so should have long names?

AnthonySteele: Both have their benefits - short when typing something quick, long for scripts, but it's not such a  big deal.

PR #83, opened 2017-07-26
Extract the nuget path logic and put a test around it

AnthonySteele:
Extract the nuget path logic and put a test around it, because it's dodgy and needs something more.
The test won't work on linux but neither does the profile path. 

So we can only do `nuget.exe` updates on windows, but it's for old-style projects that only run on windows anyway

### DIFF
--- a/NuKeeper/NuGet/Process/NugetPath.cs
+++ b/NuKeeper/NuGet/Process/NugetPath.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
 
 namespace NuKeeper.Nuget.Process
 {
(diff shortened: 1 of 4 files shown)


NuKeeper/NuGet/Process/NugetPath.cs
```diff
@@ -0,0 +1,38 @@
...
+        public static string Find()
+        {
+            var profile = Environment.GetEnvironmentVariable("userprofile");
+
+            if (string.IsNullOrWhiteSpace(profile))
+            {
+                throw new Exception("Could not find user profile path");
+            }
+
+            var commandlinePackageDir = Path.Combine(profile, ".nuget\\packages\\nuget.commandline");
```
cohen990: So obviously this is a pretty hard assumption/dependency.  Is it possible to include the nuget command line as a nuget package and refer to the one within the project?

I'm aware the dependency has existed for a little while but still

cohen990: https://www.nuget.org/packages/NuGet.CommandLine/

cohen990: I have opened #84

PR #88, opened 2017-07-28
More reliable in getting the user name

AnthonySteele:
More reliable in getting the user name from github using the token.

It seem that building the url from 2 parts only works just right if the first part ends with a trailing `/` and the second one doesn't start with one.

if the second part starts with a slash, it is taken as an absolute path, .i.e. `https://github.myco.com/api/v3` and `/user` is `https://github.myco.com/user` which is wrong.

it is unfortunate that we don't have integration tests on `GithubClient`, but that would require committing a github api token, which github disallows.


slang25:
Good job on getting to the bottom of this

Issue #86, opened 2017-07-27
Internal repo fails with 401


AnthonySteele:
On a run on an internal repo I get the following
````
Git url: https://github.je-myco.com/team/foo.git
request failed with status code: 401
````

Seems to be since the change to `lib2gitsharp`

cohen990:
using a token for the right endpoint?

PR #89, opened 2017-07-28
Supply creds when cloning

AnthonySteele:
In the git driver, Supply creds when cloning
as well as on the push
this fixes the "401" issue with some repos - #86


slang25:
Again thanks, I wish I had tested with GHE internally

PR #91, opened 2017-07-30
Use Octokit

slang25:
Wasn't quite sure how much of the old abstraction to take out.
If using a GHE address, the base url is the address you browser to. Before we needed to put in the api endpoint address.


AnthonySteele:
Nice! I will try it out later today.

AnthonySteele:
I am trying this out on our test repo ( https://github.com/NuKeeperDotNet/JustEat.StatsD ) and I get a failure when the PR is made:
```
Making PR on 'https://api.github.com/ NuKeeperDotNet JustEat.StatsD'
Update failed: NotFoundException Not Found
```

I assume that the API base url should be `https://github.com/` ?

slang25:
For github.com you should use `api=https://api.github.com/`
I had a bug in there that failed to create the PR, should be good now, tested on GH + GHE.

AnthonySteele:
Confirmed, it generates this PR: https://github.com/NuKeeperDotNet/JustEat.StatsD/pull/11

Issue #72, opened 2017-07-24
Cannot determine the packages folder to restore NuGet packages.


AnthonySteele:
On a VS2015 project,  [this line](https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper/NuGet/Process/PackagesConfigUpdater.cs#L24) gives an error message: `Cannot determine the packages folder to restore NuGet packages. Please specify either -PackagesDirectory or -SolutionDirectory.`

It works if you run `nuget restore` in the same folder as the `.sln` file instead. Doing this once may be a better option.

AnthonySteele:
It's still an open question if this is in fact an issue with the affected projects.

See https://docs.microsoft.com/en-us/nuget/consume-packages/package-restore#automatic-restore-in-visual-studio

AnthonySteele:
Still an issue with some of our repos. We get:
````
Cannot determine the packages folder to restore NuGet packages.
 Please specify either -PackagesDirectory or -SolutionDirectory.
Unable to locate the packages folder. 
Verify all the packages are restored before running 'nuget.exe update'.
````

a `nuget restore` in the same folder as the solution file does the trick.

AnthonySteele:
Finally fixed by #135

PR #92, opened 2017-07-30
Solution restore

AnthonySteele:
Fix for #72
Find all the solution files and issue a `nuget restore` or `dotnet restore` against them.
Could do with review
In general, the testability is poor.

Don't merge yet, some more investigation required
- it might be that project restore should work 
- you can have a project without a solution, in which case you should use project restore.

See https://docs.microsoft.com/en-us/nuget/consume-packages/package-restore#automatic-restore-in-visual-studio


NuKeeper/RepositoryInspection/SolutionsRestore.cs
```diff
@@ -0,0 +1,25 @@
...
+            _solutionRestoreCommand = solutionRestoreCommand;
+        }
+
+        public async Task Restore(string rootDirectory)
+        {
+            var solutionFiles = SolutionFinder.FindSolutions(rootDirectory);
+
+            foreach (var solutionPath in solutionFiles)
+            {
+                await _solutionRestoreCommand.Restore(solutionPath);
```
cohen990: I would expect to read `_command.Invoke(solutionPath);` here - doing it this way leaves a lot of openings for bundling up many different commands into this ISolutionRestoreCommand object, which would be an ugly solution

NuKeeper/Engine/RepositoryUpdater.cs
```diff
@@ -69,6 +72,23 @@ public class RepositoryUpdater
             Console.WriteLine("Done");
         }
 
+        private async Task RestoreSolutions(IEnumerable<PackageUpdateSet> updates)
```
cohen990: This feels extractable into a separate class (`ISolutionManager.RestorePackages(update)`) which would make it more testable

NuKeeper/RepositoryInspection/SolutionsRestore.cs
AnthonySteele: In general, should the things in `NuKeeper/NuGet/Process` that use `IExternalProcess` should have a "command" and "invoke" naming convention? It might be better.

AnthonySteele:
In general, tests and testability are an issue, which means isolating large parts of the app that deal with file system, git or GitHub api.

NuKeeper/RepositoryInspection/SolutionsRestore.cs
cohen990: I think that would be quite neat (but I'm kinda obsessed with Command/Invoke conventions right now... not sure why)

cohen990:
Not necessarily... can always (and probably should) require the developer to grab their own access tokens for testing - we can also dump a token into environment variables in travis to run the tests over there

PR #94, opened 2017-07-31
SimpleInjector

cohen990:


cohen990:
This may not hit every place where we are newing up a thing, but it at least gives us a point to migrate towards

PR #95, opened 2017-07-31
Don't assume "master" is the default branch

skolima:


NuKeeper/Engine/RepositoryUpdater.cs
```diff
@@ -85,7 +85,9 @@ private void GitCloneToTempDir()
             {
                 EngineReport.OldVersionsToBeUpdated(updateSet);
 
-                _git.Checkout("master");
+                var defaultBranch = _git.GetCurrentHead();
```
AnthonySteele: We should read the default branch once, before the loop on lines 62-65. 
can store it in a private member?

PR #96, opened 2017-08-01
Command pattern

AnthonySteele:
Command pattern refactor for `ExternalProcess` and things that use it.
Push a few lines of common code to check result into `ExternalProcess`.  
Things that run external processes are named as `Command` with an `Invoke` method.


NuKeeper/Engine/RepositoryUpdater.cs
```diff
@@ -119,11 +119,11 @@ private void GitCloneToTempDir()
         {
             foreach (var current in updateSet.CurrentPackages)
             {
-                var updater = current.Path.PackageReferenceType == PackageReferenceType.ProjectFile
-                    ? (INuGetUpdater) new NuGetUpdater()
-                    : new PackagesConfigUpdater();
+                var updateCommand = current.Path.PackageReferenceType == PackageReferenceType.ProjectFile
+                    ? (IUpdatePackageCommand) new DotNetUpdatePackageCommand()
```
cohen990: what's with the hard cast?

cohen990: This is one way to remove it - https://github.com/cohen990/NuKeeper/commit/fa708b843c1d24ff1c2365e1df477229e0b06d72

but I guess there's an argument for this hard cast being a reasonable choice given the situation

AnthonySteele: The hard cast is needed to get the correct type out of the `? :` operator. But I really don't mind which way, I can change it to remove the `? :`.

PR #99, opened 2017-08-02
Refactoring the program

AnthonySteele:
Extracted `Registry` and `Runner` classes from the program.
The registry is the IoC config part.
The runner is the async part, and all it's args turn into constructor params, which can be found by using the container :)
Is naming ok? "Runner" is a fairly generic name.


NuKeeper/Registry.cs
```diff
@@ -0,0 +1,26 @@
+﻿using NuKeeper.Configuration;
+using NuKeeper.Engine;
+using NuKeeper.Github;
+using NuKeeper.NuGet.Api;
+using SimpleInjector;
+
+namespace NuKeeper
+{
+    public static class Registry
```
cohen990: ContainerRegistration.Init or IOC.RegisterContainer seems more appropriate - Registry makes me think of operating systems and RAM

NuKeeper/Runner.cs
```diff
@@ -0,0 +1,56 @@
...
+            _github = github;
+            _githubToken = settings.GithubToken;
+        }
+
+        public async Task Run(string tempDir)
+        {
+            var githubUser = await _github.GetCurrentUser();
+            Console.WriteLine($"Read github user '{githubUser}'");
+
+            var git = new LibGit2SharpDriver(tempDir, githubUser, _githubToken);
```
cohen990: can we modify things so this is injectable?

NuKeeper/Program.cs
```diff
@@ -23,70 +23,16 @@ public static int Main(string[] args)
...
             var tempDir = TempFiles.MakeUniqueTemporaryPath();
 
-            RunAll(container.GetInstance<IGithubRepositoryDiscovery>(),
-                container.GetInstance<IPackageUpdatesLookup>(),
-                container.GetInstance<IPackageUpdateSelection>(),
-                container.GetInstance<IGithub>(),
-                tempDir,
-                container.GetInstance<Settings>().GithubToken)
+            var runner = container.GetInstance<Runner>();
+            runner.Run(tempDir)
```
cohen990: I always liked Engine.Run(...) just for aesthetic reasons :P

NuKeeper/Runner.cs
AnthonySteele: I expect that most or all  of this will become injectable, but in later PRs

PR #101, opened 2017-08-06
Injectable logger

AnthonySteele:
As discussed, we don't want to use scattered `Console.WriteLine` for logging.
Injectable logger class handles this.
The interface is called `INuKeeperLogger` to stop the name clash with `ILogger` in Nuget libs.
We have 3 log levels for verbosity control (Summary, Info, Verbose). and it's read from settings.
Haven't replaced every occurrence of `Console.WriteLine` yet, that will  require making some classes not static.


NuKeeper.Integration.Tests/NuGet/Api/NullNuGetLogger.cs
```diff
@@ -0,0 +1,39 @@
+using NuGet.Common;
+
+namespace NuKeeper.Integration.Tests.Nuget.Api
+{
+    public class NullNuGetLogger : ILogger
```
cohen990: should be internal

NuKeeper.Integration.Tests/NuGet/Api/NullNuKeeperLogger.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using NuKeeper.Logging;
+
+namespace NuKeeper.Integration.Tests.Nuget.Api
+{
+    public class NullNuKeeperLogger : INuKeeperLogger
```
cohen990: should be internal

PR #102, opened 2017-08-08
Split terse/summary from details

AnthonySteele:
Split terse/summary from details in some of these logging messages

### DIFF
--- a/NuKeeper/Engine/EngineReport.cs
+++ b/NuKeeper/Engine/EngineReport.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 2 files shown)


NuKeeper/Engine/EngineReport.cs
```diff
@@ -8,27 +7,39 @@ namespace NuKeeper.Engine
...
                 .Count();
 
             var packageIds = packages
+                .Select(p => p.Id)
+                .Distinct();
+
+            return $"Found {packages.Count} packages in use, {packageIds.Count()} distinct, in {projectPathCount} projects.";
+        }
+
+        public static string PackagesFoundDetails(List<PackageInProject> packages)
```
cohen990: something about this is making me think we should design it more like a proper UI layer rather than just piggybacking on the logger.

AnthonySteele: Hm, it could return a tuple or struct containing 2 strings, the summary and detail?

cohen990: Tuples are just substitute objects haha

AnthonySteele: Yeah, a real object would be better

PR #114, opened 2017-08-18
Configurable nuget sources

skolima:
The `dotnet` version now also does a restore first, because `dotnet add` only accepts one target repository name and would fail if dependencies for the update are missing.


README.md
```diff
@@ -90,7 +92,7 @@ It is similar to [this rule of preferring to use a parameter of a base type or i
 
 ## Limitations
 
-You will need command line versions of `git` and `dotnet` installed.
+You will need command line version of `dotnet` installed.
 
 It currently only runs on windows due to using `cmd` to invoke command-line processes for `git` and `dotnet`.
```
AnthonySteele: Also remove mention of git here.

AnthonySteele:
👍 Looks ok and we need this feature, many orgs will have internal nuget feeds serving packages that need auto-updating.

NuKeeper/NuGet/Api/PackageSearchMedatadataWithSource.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public string Summary => _originalMetadata.Summary;
+
+        public string Tags => _originalMetadata.Tags;
+
+        public string Title => _originalMetadata.Title;
+
+        public bool IsListed => _originalMetadata.IsListed;
+    }
+}
```
AnthonySteele: Do we need *all* of these properties surfaced? I would leave most of them off until if/when we need them.

skolima: TBH, I've implemented the full NuGet interface here, as it made for a much easier refactoring - and the implementation is autogenerated by R#. If you want, I can remove the interface and just trim it to what we actually use.

AnthonySteele: I would remove the interface and trim it down, yes.

PR #117, opened 2017-08-20
Check if the branch exists

AnthonySteele:
A useful feature not a like-for-like refactoring, for once.
This PR addresses this problem:  if you run NuKeeper, it might generate a PR e.g.
title: `Automatic update of Microsoft.NETCore.Platforms to 2.0.0`
Branches: 	NuKeeperBot  wants to merge 1 commit into `master` from `nukeeper-update-Microsoft.NETCore.Platforms-to-2.0.0` - note that the "from" branch name is deterministic.

Run NuKeeper again and it fails
`Update failed NonFastForwardException : cannot push non-fastforwardable reference`
because it tries to make the same named branch again.
So check if the target branch name exists, and if it does,  fail fast and informatively.


skolima:
TBH, I _prefer_ the current failure scenario. Consider having NuKeeper running on a daily schedule and getting a bit behind with merges/testing. Currently, you'll be  possibly missing out on the new updates - but with this change, you'll get the same update multiple times, in separate PRs.

AnthonySteele:
Then it should perhaps filter these potential PRs out earlier, i.e. "I can see a branch with the appropriate name, thereofor I won't try making this PR again" ?

skolima:
👍 to that idea

AnthonySteele:
In other words if you have `maxpr=1` and you run once daily, you have a budget on 1 PR per day.  On the second you it should try making a different PR instead of repeating the first one (regardless of success or failure).

NuKeeper/Git/LibGit2SharpDriver.cs
```diff
@@ -58,6 +58,16 @@ public void CheckoutNewBranch(string branchName)
             }
         }
 
+        public bool BranchExists(string branchName)
+        {
+            using (var repo = MakeRepo())
+            {
+                var branchFound = repo.Branches.Any(
+                    br => string.Equals(br.FriendlyName, branchName, StringComparison.OrdinalIgnoreCase));
```
skolima: git branch names are case-sensitive, this might cause issues 😅

PR #118, opened 2017-08-21
Filter out existing branches

AnthonySteele:
As discussed here https://github.com/NuKeeperDotNet/NuKeeper/pull/117
You can't make the same branch twice, so it was failing on a second run.
Instead of trying the same update with a different branch name,  make a different update.
When running again, try to make different update PRs
based on noticing that the Update's PR branch already exists.


AnthonySteele:
Do not merge, needs rework: tests and stuff.

AnthonySteele:
Looks good to me. Do we need to even mention or give an example of an `foo or bar` regex? 
The choice of input syntax here was regex or list of strings, and regex was chosen as being more flexible, i.e. can do the simple list of strings that we anticipate, and more.

Issue #126, opened 2017-08-30
Flag or mode to only do minor updates


AnthonySteele:
Feature request: 
A mode where it doesn't do major version updates, only minor and patch. On the internal feed especially, major version updates are best not automated, but patches are always useful.

AnthonySteele:
I have been prototyping this, with an enum `enum VersionChange { Patch, Minor, Major }` to say what kinds of change are allowed.

I think this will be useful for internal code, where minor or patch updates are more numerous and safer to apply than major ones.

e.g. if you are using `Newtonsoft.Json 8.0.2` and ask for only `Patch` updates, you will get an update to version `8.0.3`. 
For `Minor` you also get `8.0.3` (you would get a `8.1.0` if there was one, but there is not).
For Major you would get the latest overall, currently  `10.0.3`.

In prototyping, there are a complexities: 

* You are no longer finding latest of `Newtonsoft.Json`, you are finding an appropriate update to  `Newtonsoft.Json 8.0.2`, identified by name and version. This means passing more data around.

* What happens when you ask for `Minor` updates only, and one project uses  `Newtonsoft.Json 8.0.2` and another  `Newtonsoft.Json `10.0.2`. It is possible to give those two projects different minor version updates, it is also possible to apply a rule of "find the latest version in use, and apply minor updates to that". projects using different version in the same solution is not a good idea, IMHO it's worth just removing it wherever it's found.

cohen990:
I'd be worried about users setting their regular build to only look for minor updates and therefore the user bring unaware of major version upgrades that they can use.

Some kind of warning could be useful? "There is a major version update available" kind of message

PR #145, opened 2017-10-04
Control allowed Version change

AnthonySteele:
The engine work  for #126 is in this PR. And tests.
Not yet surfaced in a commandline flag.

we have the ` enum VersionChange { None = 0, Patch, Minor, Major }` Which controls which level of update should be selected to be applied.  `ApiPackageLookup` applies this by using the `VersionChangeFilter`.
There are sufficient tests.

No way to use this from the commandline, yet. That can be a different PR.


AnthonySteele:
By "specifying all enum values explicitly" you mean `Patch = 1` etc?

skolima:
Indeed. It's redundant, until someone deletes a value and everything shifts. I just grew to consider being explicit the lesser of two evils.

PR #148, opened 2017-10-16
Accept .vbproj as valid VS projects for PackageReference search

skolima:


AnthonySteele:
I know nothing about the latest `.vbproj` format - does it work the same way as `.csproj` as far as we are concerned?

skolima:
It's exactly the same thing (and has been pretty much since MSBuild started) - those extensions are only by convention.

PR #160, opened 2017-11-01
Handle VS 2017 projects for .NET classic

skolima:
Test uses a sample from 4.6.1 VS2017 project format, with a namespace and slightly different version storage (node instead of attribute).


AnthonySteele:
There seem to be 2 things there - 
An xml namespace  at the root node, and that the package's Version can be an element not an attr?

skolima:
Turns out the updater needs to be changed slightly as well, otherwise NuKeeper will put two versions of the dependency in.

skolima:
Relevant https://github.com/NuGet/Home/issues/4358 https://github.com/dotnet/cli/issues/5425

skolima:
Workaround used for `dotnet add package` not removing old version every time: run `dotnet remove package` first.

PR #173, opened 2018-01-01
No need to multitarget

AnthonySteele:
There doesn't seem to be a need to multitarget
As far as I know now, if you want to produce a version for net461, you can do that afterwards with `dotnet publish`

https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish

### DIFF
--- a/NuKeeper.Integration.Tests/NuKeeper.Integration.Tests.csproj
+++ b/NuKeeper.Integration.Tests/NuKeeper.Integration.Tests.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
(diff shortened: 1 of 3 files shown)


NuKeeper.Integration.Tests/NuKeeper.Integration.Tests.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
```
slang25: You might want to change this to `TargetFramework` (no "s") as this is sort of multi-targeting but for 1 framework if you get me.

AnthonySteele: Yep, that would explain why the VS tooling hadn't picked it up yet. An avoidable syntax complexity :(

PR #180, opened 2018-01-19
Fork control on commandline

AnthonySteele:
An option to prefer the upstream over using or creating a fork
comes in the command line and controls the behaviour of `ForkFinder`.

WIP

It needs input on naming, right now is

````
 public enum ForkMode
    {
        PreferFork,
        PreferUpstream
    }
````

And input on defaults, and what should happen if the first preference is not usable - e.g.  we `PreferUpstream` but on this repo we cannot push there. I don't want to make lots of options (.e.g. `PreferUpstream`, `OnlyUpstream`, `UseForkIfItExistsButDontCreateFork` etc ) unless necessary

### DIFF
--- a/NuKeeper.Tests/Configuration/SettingsParserCommandlineTests.cs
+++ b/NuKeeper.Tests/Configuration/SettingsParserCommandlineTests.cs
@@ -34,7 +34,17 @@ namespace NuKeeper.Tests.Configuration
         }
 
         [Test]
-        public void ValidRepoCommandLineHasDefaultValues()
+        public void ValidRepoCommandLineHasDefaultGithubSettings()
+        {
+            var commandLine = ValidRepoCommandLine();
+            var settings = SettingsParser.ReadSettings(commandLine);
+
+            AssertSettingsNotNull(settings);
+            Assert.That(settings.GithubAuthSettings.ApiBase, Is.EqualTo(new Uri("https://api.github.com/")));
+        }
+
+        [Test]
+        public void ValidRepoCommandLineHasDefaultUserSettings()
         {
             var commandLine = ValidRepoCommandLine();
             var settings = SettingsParser.ReadSettings(commandLine);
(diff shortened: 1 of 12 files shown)


AnthonySteele:
There are a number of options of what branches to prefer and in what order. e.g.
 * Use the user's fork if it exists and is suitable
 * Use the user's fork if it exists and is suitable, fall back to upstream
 * Use the user's fork if it exists and is suitable, create the fork if it does not, fall back to upstream
 * Use the upstream  if it exists and is suitable
 * Use the upstream  if it exists and is suitable, fall back to the  user's fork

etc.

I don't want a long list of modes, as this makes it hard to test and I'm not convinced that they're all useful.

I am happy with 
*  Use the user's fork if it exists and is suitable, create the fork if it does not, fall back to upstream

Happy to ad another mode that prefers the upstream.

skolima:
From all those options, only two so far I've seen used in practice, to be honest.
 - `PreferFork` which is how you (and most open source) work
 - `PreferUpstream` which is how most smaller organisations seem to use GitHub

And if this is a preference, with a fall back, then it kind-of also covers other more exotic scenarios. I'd be happy with having just those two options.

AnthonySteele:
Any thoughts on naming?

skolima:
I'm OK with your choice - you might want a second native speaker's input, though :-)

skolima:
I've tested this branch and realised that how I work currently is actually not what I assumed. There're several huge, archived, read-only repositories in the organisation, so the `PreferSingleRepository` would be an unacceptable performance regression (this brings the data download for the org from about 1GB to 10GB or more). Would you prefer a third mode - `SingleRepositoryOnly` or for this to replace `PreferSingleRepository`?

AnthonySteele:
So basically, you need the mode where it fails if it can't work on the main repo?

skolima:
Previously I relived on the fact it _skipped_ the repos (reporting to the stdout) it didn't have push rights to - didn't even try to clone them.

AnthonySteele:
yeah, I think you need `SingleRepositoryOnly`, plus another check of the failure handling to make sure that it carries on to the next repo.

Issue #181, opened 2018-01-24
Reporting on updates


AnthonySteele:
For internal reporting, we would like to add number and degree of package updates available to some of our internal code scorecards.  NuKeeper generates this info in detail, and then selects some updates to perform, and discards the info about the rest.

I would not like to integrate these systems closely, but NuKeeper could optionally emit this data as a report (maybe CSV) listing all the potential updates with package name, current and available versions, number of affected projects etc.

The scorecard software can parse this and generate a score from it, based in it's own weighting of criteria such as number of potential package updates, number of major version updates etc.

AnthonySteele:
output of a test version looks like this

| Package id                     | Usage count | Versions in use | Lowest version in use |  Highest Version in use | Highest available | Selected update | Package source                      | 
|--------------------------------|-------------|-----------------|-----------------------|-------------------------|-------------------|-----------------|-------------------------------------| 
| Microsoft.NETCore.Platforms    | 1           | 1               | 1.0.2                 | 1.0.2                   | 2.0.1             | 2.0.1           | https://api.nuget.org/v3/index.json | 
| System.Diagnostics.TraceSource | 1           | 1               | 4.0.0                 | 4.0.0                   | 4.3.0             | 4.3.0           | https://api.nuget.org/v3/index.json | 
| System.Net.NameResolution      | 1           | 1               | 4.0.0                 | 4.0.0                   | 4.3.0             | 4.3.0           | https://api.nuget.org/v3/index.json | 
| Microsoft.NET.Test.Sdk         | 1           | 1               | 15.0.0                | 15.0.0                  | 15.5.0            | 15.5.0          | https://api.nuget.org/v3/index.json | 
| Moq                            | 1           | 1               | 4.7.1                 | 4.7.1                   | 4.8.1             | 4.8.1           | https://api.nuget.org/v3/index.json | 
| Shouldly                       | 1           | 1               | 2.8.2                 | 2.8.2                   | 3.0.0             | 3.0.0           | https://api.nuget.org/v3/index.json | 
| xunit                          | 1           | 1               | 2.2.0                 | 2.2.0                   | 2.3.1             | 2.3.1           | https://api.nuget.org/v3/index.json | 
| xunit.runner.visualstudio      | 1           | 1               | 2.2.0                 | 2.2.0                   | 2.3.1             | 2.3.1           | https://api.nuget.org/v3/index.json |

PR #184, opened 2018-01-29
Report gen

AnthonySteele:
Generate CSV output of all possible updates
for #181

### DIFF
--- a/NuKeeper.Tests/Engine/CommitReportTests.cs
+++ b/NuKeeper.Tests/Engine/CommitReportTests.cs
@@ -218,7 +218,7 @@ namespace NuKeeper.Tests.Engine
         private static PackageUpdateSet UpdateSetFor(params PackageInProject[] packages)
         {
             var newPackage = NewPackageFooBar123();
-            return new PackageUpdateSet(newPackage, 
+            return new PackageUpdateSet(newPackage,
                 "someSource",
                 newPackage.Version,
                 VersionChange.Major,
--- a/NuKeeper/ContainerRegistration.cs
+++ b/NuKeeper/ContainerRegistration.cs
@@ -2,6 +2,7 @@
 using NuKeeper.Configuration;
 using NuKeeper.Engine;
 using NuKeeper.Engine.Packages;
+using NuKeeper.Engine.Report;
 using NuKeeper.Files;
 using NuKeeper.Github;
 using NuKeeper.Logging;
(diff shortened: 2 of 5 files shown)


zaccharles:
The report looks good, but I wonder if it can be smarter by considering which version is referenced by the entry project. I mean, which version is actually going to go out with a build of the solution. This gets more complicated when the reference comes via a project reference, but then I guess it's just about building a tree. Not sure if assembly redirects need to be considered.

It's not ideal to have old, potentially buggy or otherwise bad versions in a solution. However, if it's not actually going to go out with the rest of the code, it's less of a problem. Do you think this distinction is worth making?

AnthonySteele:
Nukeeper is purely a source code reading tool, it has no notion of which is the entry project.

In this report, the column "Usage count" says in how many projects of the solution the package is referenced.  It's possible to use a different version in different projects, so the number in the column "Versions in use" counts how many different versions of the package are used across the solution - anything over one is bad. 

Because of this, it's not always possible to give one number for the "version in use", that's why the next two columns show the lowest and highest version in use. These will always be the same if "Versions in use"  count is one. Yes, it's the one in the entry project that really matters as it overrides the rest at build, but it might not during tests. NuKeeper does aim to reduce these "conflicts" as a priority over other package updates.  IMHO the whole thing becomes more reasonable if you eliminate them.

There are also packages (e.g. `Moq`, `NUnit`) that get used in the solution but not in the deployed project.  NuKeeper makes no distinction here, as they also benefit from updates.

The real test of what is deployed is for tools that look at the output of the build process, but unfortunately then you are looking at `.dll` files and not nuget packages, and there isn't a one-to-one mapping between them.

zaccharles:
Understood. I think this report is good as far as what NuKeeper knows about and can do at the moment. For a future integration with Score Card (obviously a separate concern), we should just consider what kind of value we can get out of it.

AnthonySteele:
The simplest metric to get from that report is the number of lines in it. Each line past the heading is a potential update, and fewer is better.

zaccharles:
I think we can do better, though. For example, by considering patch, minor, and major version differences with a different priority. Considering how many versions behind (20 version is worse than 1 version?). Certainly how many packages, though even that is questionable as it punishes big projects more than smaller projects. It all just needs to boil down to a number between 1 and 100, and a list of possible actions/problems, links to various things, and/or a link out to a report (it supports markdown and a simple key/value grid where a value can be a markdown string or another grid). 
Will just need a bit of thinking.

AnthonySteele:
Right, the report does not include all the data for major, minor and patch updates yet and I'm still thinking about how to keep this data and show it.
e.g. 
you are using Newtonsoft.Json `10.0.1`, you can update to
Patch: `10.0.3`
Minor: none, or  `10.0.3` if you consider this inclusive of lesser updates
Major: none, or  `10.0.3` 

you are using Newtonsoft.Json `8.0.2`, you can update to
Patch: `8.0.3`
Minor: none, or  `8.0.3`  if you consider this inclusive of lesser updates
Major: `10.0.3`

PR #188, opened 2018-02-08
Updating to 2.1.4 dotnet core sdk

skolima:


skolima:
This raises an interesting question - should those SDK upgrade be also automatically flagged and PRs created?

PR #211, opened 2018-03-04
A better package update sort

AnthonySteele:
A better package update sort.

**Previously**: sorted only on the number of different versions in use.

**Now**: sort on the number of different versions in use, then the number of projects affected, then on the largest version change.
Older updates get a score boost. Change from pre-release to not pre-release gets a score boost.

And tests on it

### DIFF
--- a/NuKeeper.Tests/Engine/Packages/PackageUpdateSortTests.cs
+++ b/NuKeeper.Tests/Engine/Packages/PackageUpdateSortTests.cs
@@ -29,7 +29,7 @@ namespace NuKeeper.Tests.Engine.Packages
         {
             var items = new List<PackageUpdateSet>
             {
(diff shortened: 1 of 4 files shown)


AnthonySteele:
Does the sort cache scores? It seems slow, and the `Priority` function might be expensive.

AnthonySteele:
Maybe for a later PR, but I was trialing this code and noticed that this PR was _not_ sorted at the top:
`NuKeeper has generated an update of Amazon.Lambda.Tools to 1.9.0 from 1.3.0-preview1`

Changes that get you out of beta should have a higher score.

AnthonySteele:
It just needs a test on "when updates have different  published dates, the oldest is preferred"

PR #213, opened 2018-03-07
Read user identity from the GitHub token

skolima:
Prevents ArgumentNullException when no .gitconfig exists


NuKeeper/Git/LibGit2SharpDriver.cs
```diff
@@ -99,7 +101,7 @@ public void Commit(string message)
             _logger.Verbose($"Git commit with message '{message}'");
             using (var repo = MakeRepo())
             {
-                var sig = repo.Config.BuildSignature(DateTimeOffset.Now);
```
AnthonySteele: This looks like an oversight, well found!

PR #219, opened 2018-03-14
Fallback to .gitconfig if GitHub profile information is missing

skolima:
Add output explaining what is going on.

### DIFF
--- a/NuKeeper/Engine/GithubEngine.cs
+++ b/NuKeeper/Engine/GithubEngine.cs
@@ -1,8 +1,11 @@
+using System;
 using System.Threading.Tasks;
 using LibGit2Sharp;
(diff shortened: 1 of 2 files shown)


NuKeeper/Engine/GithubEngine.cs
```diff
@@ -47,5 +54,19 @@ public class GithubEngine
                 await _repositoryEngine.Run(repository, gitCreds, userIdentity);
             }
         }
+
+        private Identity GetUserIdentity(Account githubUser)
+        {
+            try
```
AnthonySteele: I would return null right away if `string.IsNullOrWhiteSpace(githubUser?.Name)` or the Email, and avoid throwing and catchingthe exception in that case.

Issue #127, opened 2017-08-30
Full netcore support on linux


AnthonySteele:
The all-netcore scenario should work:
Updating netcore solutions should work on Linux and mac machines

AnthonySteele:
Linux / amc can be made to work, the issues are
 * file system operations and tests need to be coded to not assume dir separator chars etc 
 * We use external os processes to run `nuget.exe` or `dotnet.exe`. Need a version of the `ExternalProcess` class for linux, which can then run the `dotnet` executable on that system.

Other things in use, like `octokit` and `libgit`, should already work fine.

AnthonySteele:
It works now!

PR #227, opened 2018-03-22
Path handling behaviour on Linux

skolima:
following #127

### DIFF
--- a/NuKeeper.Tests/Engine/Packages/PackageUpdateSortTests.cs
+++ b/NuKeeper.Tests/Engine/Packages/PackageUpdateSortTests.cs
@@ -8,7 +8,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Runtime.InteropServices;
 
 namespace NuKeeper.Tests.Engine.Packages
 {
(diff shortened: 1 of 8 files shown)


NuKeeper.Tests/Engine/Packages/PackageUpdateSortTests.cs
```diff
@@ -210,7 +212,10 @@ private static PackageUpdateSet UpdateSetFor(PackageIdentity package, DateTimeOf
 
         private static PackageInProject MakePackageInProjectFor(PackageIdentity package)
         {
-            var path = new PackagePath("c:\\temp", "folder\\src\\project1\\packages.config",
+            var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            var baseDirectory = isWindows ? "c:\\temp" : "/temp";
```
AnthonySteele: I think it's worth pulling these 2 lines out into a new static class & method, as they occur several times below, and they control how you interact with the OS.

AnthonySteele: Hm, `c:\temp` will Work On My Machine and that's about it. Should be `Path.GetTempPath` `https://msdn.microsoft.com/en-us/library/system.io.path.gettemppath(v=vs.110).aspx`

skolima: Ha, those are unit tests. This doesn't touch the actual file system, I have not `c:\temp` and the tests still work - they're building paths in memory. So, that's a starting point for parsing, not for actual file system access. Perhaps this could be more explicit by using `x:\imaginary-folder`?

AnthonySteele: Ah, fine then.

PR #228, opened 2018-03-25
Better branch filter

AnthonySteele:
Replace the git check for an existing local branch,  as it's no longer useful, with one that works. 

It should now be an inspection of the push fork to see if it already has an existing remote branch with that name, and move on to other updates in that case.

This is surprisingly not at all simple, as the old check was sync and the new check is async.  Each package checks for an existing branch on github, and these are async operations. You need to pass data around more, need new methods and an async list Filter.

### DIFF
--- a/NuKeeper.Tests/Engine/PackageUpdateSelectionTests.cs
+++ b/NuKeeper.Tests/Engine/PackageUpdateSelectionTests.cs
@@ -24,7 +24,7 @@ namespace NuKeeper.Tests.Engine
 
             var target = OneTargetSelection();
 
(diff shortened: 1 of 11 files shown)


AnthonySteele:
Hard to test :( as the `OcotkitClient` requires github creds to work, and you should not and cannot check working github creds into a github repo. 

Proof of it working is that two runs produce 2 different PRs ( https://github.com/NuKeeperDotNet/JustEat.StatsD/pulls ) as the second run detects the branch made by the first run, and doesn't try (and fail) to make it again.

PR #237, opened 2018-04-06
Update Assembly Binding redirects

AnthonySteele:


skolima:
Uhm, why regex instead of xpath? That would be somewhat less fragile when it comes to whitespace and formatting (and seems to be how NuGet itself applies the bindings).

I'd need to check deeper into how the target binding is checked/generated, but it seems somewhat off? Newtonsoft.Json for example generates bindings to e.g. 11.0.0.0 no matter which release you use (so e.g. 11.0.0.2 also has 11.0.0.0 as target version).

skolima:
Indeed, NuGet uses xpath/xelement to locate the redirects :-) https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Runtime/BindingRedirectManager.cs#L216

Similarly for the nuspec dependencies, xelement is used https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Packaging/NuspecReader.cs#L116

AnthonySteele:
**Why regex to work with XML?**

The problem is that the template files used to generate configs that I see _aren't valid XML_ due to razor markup. So they can't be parsed as XML. 

Fortunately that meta-stuff is mostly in the `appSettings` section, and doesn't go into the `<runtime>` section.

skolima:
Hmm, but `.nuspec` and `.config` files will always be valid xml or they don't work. Perhaps then just the `.config.razor` ones should get a special regex handler, as those are JustEat-specific? I get that it's some duplication, but, well, people also tend to format their config files to their liking.

AnthonySteele:
**Target binding AssemblyVersion**

E.g. looking at `Newtonsoft.Json 10.0.3` the only place that I can find the number `10.0.0.0` is to
* Get the package `Newtonsoft.Json 10.0.3`
* Unpack it
* locate `Newtonsoft.Json.dll`
* Open it in [dotPeek](https://www.jetbrains.com/decompiler/) or similar tool (c# code reflection can do this too)
* locate the `AssemblyVersion` attribute.

So this procedure to find the correct AssemblyVersion make this feature a lot harder, if it is possible at all. 
Possibly updating `.nuspec` files should come first? They seem much easier to work with.

martincostello:
For reading the assembly version attribute, [`Mono.Cecil`](https://github.com/jbevain/cecil) is something I've used before which doesn't actually load the assemblies into a code context, because that causes loads of woe I won't go into, so you can pull the version out of the IL metadata without worrying about what framework it targets vs. what framework you're running your own code in.

AnthonySteele:
Yep. Though the hard part is all the steps before that starting with "Get the package", NuKeeper does not do that currently, though it might be done implicitly by a package version update.

Issue #236, opened 2018-04-05
Nukeeper Program organisation


AnthonySteele:
People have suggested pulling out the following parts as something easily invokable, maybe in separate projects in the NuKeeper solution.

Two scenarios:

1) 

I have code already checked out of git on my local hard drive. I want to inspect it and generate the list of potential nuget package upgrades. 

No interaction with git or github is needed, the files as they exist on the file system will be used. No nuget updates will be applied, so no files will be changed locally, just a report generated.

2) 

I have code already checked out of git on my local hard drive. I want to apply a package update locally across the solution. 

No interaction with git or github is needed, the files as they exist on the file system will be used. Nuget updates will be applied to change files locally, but no git commit or github PRs will be raised.

AnthonySteele:
Will surface some more pieces on the commandline next...

AnthonySteele:
Local `update` command does #2 now

PR #238, opened 2018-04-10
Project reorganisation

AnthonySteele:
#236 
splitting up the one code project.
The idea is for a library that can be called with
`Task<List<PackageUpdateSet>> FindPackageUpdateSets(IFolder workingFolder);` - this is the entry point of `UpdateFinder`
I.e. for a folder containing an already present codebase, list all the package updates for it.
This contains all the code for the inspection of csproj files, and for querying the nuget api for updates.
It seems a reasonable chunk to extract, and could have its own commandline interface.


skolima:
This generally looks good - a huge number of moves. Is the NuKeeper.Types used at all, or is that a remainder from earlier refactor?

I'll see if I have time to give this branch a test go today.

AnthonySteele:
`NuKeeper.Types` is left over from earlier commits, I will chase down the remnants of it.
It was for " shared pieces" but it isn't needed yet. If there's another extraction to pull out e.g. all the git / github operation code, then there will be a "third piece" to hold shared infra like loggers and formatters. But it isn't needed now.

skolima:
This branch WfM ;-)

Issue #243, opened 2018-04-13
Failure to update dotnet core projects with multiple projects in same folder


skolima:
I'm not entirely sure this is a valid scenario in dotnet core, but don't really see why it wouldn't be.

skolima:
https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper.Integration.Tests/NuGet/Process/DotNetUpdatePackageCommandTests.cs#L71 - reproduction in a test

Should be an easy fix, passing the project file name explicitly to the command should solve this.

Issue #239, opened 2018-04-12
Failure to update .NET AspNet WebApi projects


skolima:
Update run fails with the following:

> error MSB4019: The imported project "C:\Program Files\dotnet\sdk\2.1.104\Microsoft\VisualStudio\v15.0\WebApplications\Microsoft.WebApplication.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.

Other projects in the solution update correctly, this seems like a possible bug in dotnetcore? Need to check if it happens with a new AspNet WebApi .NET (classic) or is this something from my historical project leftovers.

skolima:
This works with a clean template generated from VS 2017: https://github.com/skolima/NewWebApiFromVsTemplate (although we do log errors trying to restore pacakges)
Possibly the problem is related to using `<PackageReference/>` (a new templated project still uses `packages.config` dependency specification.

skolima:
https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper.Integration.Tests/NuGet/Process/DotNetUpdatePackageCommandTests.cs#L58 - reproduction in a test.

skolima:
This would potentially work with `NuGet.exe`, except it doesn't support `<PackageReference/>` format at all - this seems to be by design https://github.com/NuGet/Home/issues/6115

PR #244, opened 2018-04-13
Updaters tests

skolima:
We're missing tests for those - I'll add at least those we know that work.

Some related issues found: #239 #243


skolima:
I'm trying to add a test case for working (well I'm using that scenario!) case with `dotnet add` against a "classic" csproj and PackageReference and I'm failing to create a working test case :-)

skolima:
@AnthonySteele I'd like to merge this - please let me know if the style is ok?

PR #250, opened 2018-04-17
More logging when filtering

AnthonySteele:
It's not good to see "Selection of package updates: 1 potential, filtered to 0" and not know which filter threw it out. Min package age, include/exclude by name rules, or existing branch check. This divides it into local rules and remote existing  branches.


skolima:
Instead of `IList<T>`, perhaps `IReadOnlyCollection<T>` is more explicit about the purpose? It's also implemented by `List<T>`, so should be no implementation change, just showing the purpose in the method signature better?

AnthonySteele:
@skolima good point, change made

PR #262, opened 2018-04-29
Sort by dependencies

AnthonySteele:
When packages have been sorted by priority weighting, we look at how they depend upon each other and resort if there are any dependencies among them.

If package `A` depends upon package `B` and both of them are in the update list, then `B` must come before `A` in the list, so that it is updated first.

The algorithm is [Depth-first Topological  sort](https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search), which isn't that hard if you copy it from wikipedia.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: csharp
 mono: none
 dist: trusty
-dotnet: 2.1.4
+dotnet: 2.1.105
 script:
   - dotnet build -c Release -f netcoreapp2.0
   - dotnet test -c Release -f netcoreapp2.0 NuKeeper.Tests/NuKeeper.Tests.csproj --filter "TestCategory!=WindowsOnly"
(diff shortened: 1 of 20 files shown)


CrispyDrone:
@AnthonySteele 

Why exactly is this required?

If we continue the example of package `A` that depends on package `B`; I can understand that by doing this you can keep each commit atomic as updating the dependency (`B`) will not result in the dependent (`A`) getting updated concomitantly (as far as I understand, see later in this text). The other way around i.e. updating the dependent (`A`) first, would probably result in updating the dependency `B` at the same time (if it requires a newer version of `B` of course).

However it seems like first upgrading package `A`, and then `B` independently of each other can easily cause a breaking change if package `B` is not adapted to be able to deal with an API change in `A`. 

Does NuGet somehow update package `A` when you only ask for an upgrade of package `B`? But if that's the case, then the sorting wouldn't have any function? See this related issue: https://github.com/NuKeeperDotNet/NuKeeper/issues/940

PR #264, opened 2018-05-02
Updating dependencies in nuspec files

skolima:
This PR adds `.nuspec` as another recognised source of package dependencies (for cases where a non-floating version number is used). Those dependencies will be updated along `<PackageReferrence/>` or `packages.config` dependencies and reported in PR messages created by NuKeeper.

I've chosen this approach as opposed to having just the updater (without the dependencies finder), as projects are likely not to have a 1-to-1 relationship between project files and `.nuspec` files, thus the stand-alone updater would have to be more complex. Also, the alternative approach wouldn't include the update in the PR description.

### DIFF
--- a/NuKeeper.Inspection/RepositoryInspection/RepositoryScanner.cs
+++ b/NuKeeper.Inspection/RepositoryInspection/RepositoryScanner.cs
@@ -9,9 +9,9 @@ namespace NuKeeper.Inspection.RepositoryInspection
     {
         private readonly IReadOnlyCollection<IPackageReferenceFinder> _finders;
 
-        public RepositoryScanner(ProjectFileReader projectFileReader, PackagesFileReader packagesFileReader)
+        public RepositoryScanner(ProjectFileReader projectFileReader, PackagesFileReader packagesFileReader, NuspecFileReader nuspecFileReader)
         {
-            _finders = new IPackageReferenceFinder[] {projectFileReader, packagesFileReader};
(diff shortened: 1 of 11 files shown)


NuKeeper/NuGet/Process/UpdateNuspecCommand.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                return;
+            }
+
+            var packageNodeList = packagesNode.Elements()
+                .Where(x => x.Name == "dependency" && x.Attributes("id").Any(a => a.Value == currentPackage.Id));
+
+            foreach (var dependencyToUpdate in packageNodeList)
+            {
+                dependencyToUpdate.Attribute("version").Value = newVersion.ToString();
```
AnthonySteele: We should log at verbose level that a package was updated in a nuspec file.

NuKeeper.Inspection/RepositoryInspection/RepositoryScanner.cs
```diff
@@ -7,56 +7,34 @@ namespace NuKeeper.Inspection.RepositoryInspection
...
+            _finders = new IPackageReferenceFinder[] {projectFileReader, packagesFileReader, nuspecFileReader};
         }
 
         public IEnumerable<PackageInProject> FindAllNuGetPackages(IFolder workingFolder)
         {
-
-            return PackageFiles(workingFolder)
-                .Concat(ProjectFiles(workingFolder))
+            return _finders
+                .SelectMany(f => FindPackages(workingFolder, f))
```
AnthonySteele: Nice

PR #268, opened 2018-05-12
Package updater - separate local from github ops

AnthonySteele:
This is step towards having those as a separate lib / separately invokable


AnthonySteele:
Agreed about naming. The `LocalPackageUpdater`  is actually doing the update, and the `PackageUpdater` is doing git checkout, update, git commit. I'll try to find better names. `LocalPackageUpdater` could be `UpdateRunner`.

AnthonySteele:
I still haven't found quite the right name for `PackageUpdater` - it's a "make branch for package update, apply update and make the PR; orchestrator" But the rename doesn't need to be in this PR.

PR #272, opened 2018-05-19
Use a Directory.Build.props file to hold common things

AnthonySteele:
Use a `Directory.Build.props` file to hold common things
Like framework version
As they should be all the same

This file is automatically imported into all projects under it: https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,16 @@
+<Project>
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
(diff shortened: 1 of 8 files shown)


AnthonySteele:
Integration tests fail after this, not sure why it makes a difference :(

AnthonySteele:
Integration tests that use the file system to test manipulating project files have to use a folder elsewhere on the file system, otherwise the `Directory.Build.props` file in a folder above affects their behaviour.

PR #273, opened 2018-05-19
Use "inspect" mode as a default

AnthonySteele:
Use `inspect` mode as a default.
Inspect mode is a suitable default as it's a read-only operation.
This is mostly a user-experience thing for newbies, as just running NuKeeper without args might now do something interesting, but not change anything.


AnthonySteele:
See #264

PR #287, opened 2018-06-04
Adds automatic labelling to created PRs.

skolima:
By default, use 'nukeeper' so that the automated PRs can be easily found. Multiple labels can be configured by config file or through command line.

### DIFF
--- a/NuKeeper/Engine/Packages/PackageUpdater.cs
+++ b/NuKeeper/Engine/Packages/PackageUpdater.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Linq;
 using System.Threading.Tasks;
+using NuKeeper.Configuration;
 using NuKeeper.Git;
 using NuKeeper.Github;
 using NuKeeper.Inspection.Logging;
@@ -13,16 +15,19 @@ namespace NuKeeper.Engine.Packages
     {
         private readonly IGithub _github;
         private readonly INuKeeperLogger _logger;
+        private readonly ModalSettings _modalSettings;
         private readonly IUpdateRunner _updateRunner;
 
         public PackageUpdater(
             IGithub github,
             IUpdateRunner localUpdater,
-            INuKeeperLogger logger)
+            INuKeeperLogger logger,
+            ModalSettings modalSettings)
         {
             _github = github;
             _updateRunner = localUpdater;
             _logger = logger;
+            _modalSettings = modalSettings;
         }
 
         public async Task MakeUpdatePullRequest(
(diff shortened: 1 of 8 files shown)


NuKeeper/Engine/Packages/PackageUpdater.cs
```diff
@@ -79,7 +84,12 @@ public class PackageUpdater : IPackageUpdater
                 Body = CommitWording.MakeCommitDetails(updates)
             };
 
-            await _github.OpenPullRequest(repository.Pull, pr);
+            var createdPullRequest = await _github.OpenPullRequest(repository.Pull, pr);
+
+            if (_modalSettings.Labels.Any())
+            {
+                await _github.AddLabelsToIssue(repository.Pull, createdPullRequest.Number, _modalSettings.Labels);
```
AnthonySteele: I would have put this `AddLabelsToIssue` call inside `OpenPullRequest`

skolima: I had it like this first and it felt a bit... Odd. As in, the REST API for GitHub actually allows it in one call, it's just OctoKit being odd and not handling this, so it makes sense to "fix" it in IGithub... but then it's also putting more "opinion" into the driver instead of, well, PackageUpdater. I can change this back to be in Octokit driver if you like it better.

AnthonySteele: I would prefer that, yes. I think the GitHub driver can be a bit opinionated, it's not totally general purpose, it just has to serve NuKeeper purposes. But it should cope with null/empty labels, I

NuKeeper/Github/OctokitClient.cs
```diff
@@ -88,17 +89,23 @@ public OctokitClient(GithubAuthSettings settings, INuKeeperLogger logger)
...
         }
 
-        public async Task AddLabelsToIssue(ForkData target, int number, string[] labels)
-        {
-            _logger.Info($"Adding label(s) '{labels.JoinWithCommas()}' to issue '{_apiBase} {target.Owner}/{target.Name} {number}'");
-            await _client.Issue.Labels.AddToIssue(target.Owner, target.Name, number, labels);
-        }
-
-        public async Task<PullRequest> OpenPullRequest(ForkData target, NewPullRequest request)
+        public async Task<PullRequest> OpenPullRequest(ForkData target, NewPullRequest request, string[] labels)
```
AnthonySteele: Can be `IEnumerable<string>` here ?

skolima: changed

NuKeeper/Github/OctokitClient.cs
```diff
@@ -88,17 +89,23 @@ public OctokitClient(GithubAuthSettings settings, INuKeeperLogger logger)
...
         {
             _logger.Info($"Making PR onto '{_apiBase} {target.Owner}/{target.Name} from {request.Head}");
             _logger.Verbose($"PR title: {request.Title}");
-            return await _client.PullRequest.Create(target.Owner, target.Name, request);
+            var createdPullRequest = await _client.PullRequest.Create(target.Owner, target.Name, request);
+
+            var labelsToApply = labels?.Where(l => !string.IsNullOrWhiteSpace(l)).ToArray();
```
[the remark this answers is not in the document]
skolima: changed

PR #290, opened 2018-06-10
Local Update mode

AnthonySteele:
`mode=update` works like `inspect` but will apply 1 update to the local file system.
Can be used with `dir=`
Update is automatically selected, as the top item by sort order.
Filters apply, so you could use `i=` to specify the package to update.


AnthonySteele:
ok, I think that this PR has gone far enough now.

Issue #278, opened 2018-05-22
Infer nuget package sources


AnthonySteele:
When NuKeepeer uses private nuget feeds or multiple feeds, we specify nuget package sources in config or commandline via `sources=http://packages.sourceA.com/feed;http://packages.sourceB.com/feed` This is one of the longest parts of the commandline.

However in many cases there is [a `nuget.config` file](https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior) containing `packageSources` data for the repo, listing these sources.

This is a good practice which should be encouraged, as it makes the build more self-contained - it works due to package source settings in the repo rather than global package source settings that may not be present on a new machine.

In this case NuKeepeer can read the sources out of this file instead of having to specify them, either by default or with a switch e.g.  `sources=configfile`

AnthonySteele:
Done!

PR #298, opened 2018-06-20
Infer nuget package sources for nuget.config

AnthonySteele:
#278 

Infer nuget package sources from the file `nuget.config` when it is present and not overridden on the commandline.

The fallback sequence is now:
 * If you specify nuget sources on the command line options, they will be used. User override wins.
 * if a `nuget.config` file is found and can be parsed for sources, those sources will be used.
 * The global feed on `nuget.org` is the fallback option. This will actually be 100% correct for many public projects, and will give some meaningful results for even internal code.

The new classes start with `NugetSourcesReader` that does the above logic.  And reader classes called by that to parse the file.

Most of the changes are because the `NuGetSources`  cannot now be determined from config and then put into the IoC container upfront. Instead, when there is a Folder containing the code, it does the `NugetSourcesReader` logic, and thereafter passes along the `NuGetSources` object. `NuGetSources` is no longer a constructor param, now it's a method param. This touches a lot of classes [in order to to pass it to where it is used](https://en.wiktionary.org/wiki/tramp_data).

### DIFF
--- a/NuKeeper.Inspection.Tests/Sources/NugetConfigFileParserTests.cs
+++ b/NuKeeper.Inspection.Tests/Sources/NugetConfigFileParserTests.cs
@@ -1,6 +1,8 @@
 using NuKeeper.Inspection.Sources;
 using NUnit.Framework;
+using System.IO;
 using System.Linq;
+using System.Text;
 
 namespace NuKeeper.Inspection.Tests.Sources
 {
@@ -14,7 +16,7 @@ namespace NuKeeper.Inspection.Tests.Sources
 
             var parser = new NugetConfigFileParser(new NullNuKeeperLogger());
 
-            var sources = parser.Parse(data);
+            var sources = parser.Parse(ToStream(data));
 
             Assert.That(sources, Is.Null);
         }
@@ -26,7 +28,7 @@ namespace NuKeeper.Inspection.Tests.Sources
 
             var parser = new NugetConfigFileParser(new NullNuKeeperLogger());
 
-            var sources = parser.Parse(data);
+            var sources = parser.Parse(ToStream(data));
 
             Assert.That(sources, Is.Null);
         }
@@ -44,7 +46,7 @@ namespace NuKeeper.Inspection.Tests.Sources
 
             var parser = new NugetConfigFileParser(new NullNuKeeperLogger());
 
-            var sources = parser.Parse(data);
+            var sources = parser.Parse(ToStream(data));
 
             Assert.That(sources, Is.Not.Null);
             Assert.That(sources.Items.Count, Is.EqualTo(1));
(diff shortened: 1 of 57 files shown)


NuKeeper.Inspection.Tests/Sources/NugetConfigFileParserTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        [Test]
+        public void ValidDataWithOneItemReturnsThatItem()
+        {
+            var data =
+@"<?xml version=""1.0"" encoding=""utf-8""?>
+<configuration>
+  <packageSources>
+    <add key=""NuGet official"" value=""https://nuget.org/api/v2/"" />
```
skolima: I know it's the default at your $work, and this is just test data, but it's a bit odd to see v2 here :D

NuKeeper.Inspection.Tests/Sources/NugetSourcesReaderTests.cs
```diff
@@ -0,0 +1,88 @@
...
+            var ff = new FolderFactory(new NullNuKeeperLogger());
+            return ff.UniqueTemporaryFolder();
+        }
+
+        private static INugetSourcesReader MakeNuGetSourcesReader()
+        {
+            var logger = new NullNuKeeperLogger();
+            return new NugetSourcesReader(
+                new NugetConfigFileReader
+                    (new NugetConfigFileParser(logger), logger), logger);
```
skolima: 😂

NuKeeper.Inspection.Tests/Sources/NugetConfigFileParserTests.cs
AnthonySteele: Yep, i should change that.

NuKeeper.Inspection.Tests/Sources/NugetSourcesReaderTests.cs
AnthonySteele: I wanted to refactor, but then I decided to keep the tests on the `NugetConfigFileParser` so I could not

AnthonySteele:
Yep, we can't really traverse the whole file system.

skolima:
Ha, you've  edited your comment so my late-night revelation isn't a revelation any more, but: came to the same conclusion :-) I was thinking about the scenarios "full" NuGet deals with, a developer having multiple levels of the `nuget.config` in their work folder and arranging projects in there - but for NuKeeper scenario, the only concern really is  a top level configuration file in the repository root.

AnthonySteele:
Yep, this should work as a repo-level config file; either in the root folder or sometimes one level down, e.g. in a `/src` folder.

Issue #307, opened 2018-07-04
Inspecting a project with  multiple target frameworks fails


skolima:
Failing project: https://github.com/tathamoddie/System.IO.Abstractions

Exception:
```Unhandled Exception: System.ArgumentException: An item with the same key has already been added. Key: NUnit
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at NuKeeper.Inspection.NuGetApi.BulkPackageLookup.ProcessLookupResult(PackageLookupResult packageLookup, Dictionary`2 result) in C:\Code\NuKeeper\NuKeeper.Inspection\NuGetApi\BulkPackageLookup.cs:line 56
   at NuKeeper.Inspection.NuGetApi.BulkPackageLookup.FindVersionUpdates(IEnumerable`1 packages, NuGetSources sources, VersionChange allowedChange) in C:\Code\NuKeeper\NuKeeper.Inspection\NuGetApi\BulkPackageLookup.cs:line 42
   at NuKeeper.Inspection.NuGetApi.PackageUpdatesLookup.FindUpdatesForPackages(IReadOnlyCollection`1 packages, NuGetSources sources, VersionChange allowedChange) in C:\Code\NuKeeper\NuKeeper.Inspection\NuGetApi\PackageUpdatesLookup.cs:line 27
   at NuKeeper.Inspection.UpdateFinder.FindPackageUpdateSets(IFolder workingFolder, NuGetSources sources, VersionChange allowedChange) in C:\Code\NuKeeper\NuKeeper.Inspection\UpdateFinder.cs:line 40
   at NuKeeper.Local.LocalEngine.GetSortedUpdates(IFolder folder, NuGetSources sources, VersionChange allowedChange) in C:\Code\NuKeeper\NuKeeper\Local\LocalEngine.cs:line 64
   at NuKeeper.Local.LocalEngine.Run(SettingsContainer settings) in C:\Code\NuKeeper\NuKeeper\Local\LocalEngine.cs:line 45
   at NuKeeper.Program.Main(String[] args) in C:\Code\NuKeeper\NuKeeper\Program.cs:line 28
   at NuKeeper.Program.<Main>(String[] args)```

skolima:
[...]

PR #309, opened 2018-07-04
Handle packages differing only by name case

skolima:
NuGet treats them as identical.

For issue #307

### DIFF
--- a/NuKeeper.Integration.Tests/NuGet/Api/BulkPackageLookupTests.cs
+++ b/NuKeeper.Integration.Tests/NuGet/Api/BulkPackageLookupTests.cs
@@ -49,6 +49,25 @@ namespace NuKeeper.Integration.Tests.NuGet.Api
         }
 
         [Test]
+        public async Task FindsSingleUpdateForPackageDifferingOnlyByCase()
+        {
+            var packages = new List<PackageIdentity>
+            {
+                Current("nunit"),
+                Current("NUnit")
+            };
+
+            var lookup = BuildBulkPackageLookup();
+
+            var results = await lookup.FindVersionUpdates(
+                packages, NuGetSources.GlobalFeed, VersionChange.Major);
+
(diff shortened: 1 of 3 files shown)


NuKeeper.Integration.Tests/NuGet/Api/BulkPackageLookupTests.cs
```diff
@@ -48,6 +48,25 @@ public class BulkPackageLookupTests
...
+
+            var lookup = BuildBulkPackageLookup();
+
+            var results = await lookup.FindVersionUpdates(
+                packages, NuGetSources.GlobalFeed, VersionChange.Major);
+
+            Assert.That(results, Is.Not.Null);
+            Assert.That(results.Count, Is.EqualTo(1));
+            Assert.That(results, Does.ContainKey("NUnit"));
+        }
```
AnthonySteele: Can you also ` Assert.That(results, Does.ContainKey("nunit"));` if that's true?

skolima: `Count` is equal to `1`, only the canonical package name from NuGet is returned. Oh, but you mean lookup now allows searching by both? Yeah, that makes sense.

skolima: Oh. This assert ignores the dictionary comparer and uses the default one. That's naughty `NUnit`!

AnthonySteele: Yes, if the key matching is case-insensitive then the dictionary should reply "yes, I have this key" for either caps.

AnthonySteele: Would ` Assert.That(results.ContainsKey("nunit", Is.True)` work?

PR #312, opened 2018-07-06
Move from archived NuGet.Protocol.Core.v3 to NuGet.Protocol 4.7.0

skolima:
There's 2 API differences:

1) `var cacheContext = new SourceCacheContext()` - this is now required, context with default options means "allow disk cache, not older than 30 minutes, update disk cache"
2) the `ILogger` interface has changed, I've updated our adapter.

Because of 2) I'd have to also update the NullLogger implementation in tests - instead I replaced the NullLoggers with NSubstitute. If that's against your personal preference, I can rewrite this part.

BTW, this PR unlocks using NuGet.Configuration for reading `nuget.config` files, and this will mean also handling the more obscure bits of source config there, like proxies and authentication.


AnthonySteele:
NullLoggers with NSubstitute if fine.

PR #314, opened 2018-07-06
 Use NuGet.Configuration own settings parsing logic

skolima:


skolima:
I think this is big enough for now - I'll check how the default (global) feed fallback behaves later and possibly follow with another PR. Is this one good to merge @AnthonySteele ?

Issue #311, opened 2018-07-06
Support PackageReference without version


hangy:
With ASP.NET Core 2.1, the main package reference for the core projects is changed from
```
<PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.8" />
```
to 
```
<PackageReference Include="Microsoft.AspNetCore.App" />
```

I think the actual version is determined by the SDK. Running NuKeeper on a csproj that contains this PackageReference displays an `ArgumentNullException`. This package could be skipped, so that the error is not shown.
```
Could not read package from <PackageReference Include="Microsoft.AspNetCore.App" /> in file C:\Users\me\Source\Repos\Foo\Bar\Bar.csproj ArgumentException : Value cannot be null or an empty string.
Parameter name: value
```

Probably caused by the `NuGetVersion` constructor at https://github.com/NuKeeperDotNet/NuKeeper/blob/ec27e14a1e8bb5b3aa02f107b887cc6b20423c1b/NuKeeper.Inspection/RepositoryInspection/PackageInProject.cs#L15

skolima:
That's... odd :-) this is still a package listed normally on NuGet  https://www.nuget.org/packages/Microsoft.AspNetCore.App/

skolima:
Simple repro: `dotnet new webapi; nukeeper log=Verbose`

call stack:
```
Could not read package from <PackageReference Include="Microsoft.AspNetCore.App" /> in file c:\carnect\test\test.csproj ArgumentException : Value cannot be null or an empty string.
Parameter name: value
   at NuGet.Versioning.NuGetVersion.Parse(String value)
   at NuGet.Versioning.NuGetVersion..ctor(String version)
   at NuKeeper.Inspection.RepositoryInspection.PackageInProject..ctor(String id, String version, PackagePath path) in C:\Code\NuKeeper\NuKeeper.Inspection\RepositoryInspection\PackageInProject.cs:line 15
   at NuKeeper.Inspection.RepositoryInspection.ProjectFileReader.XmlToPackage(XElement el, PackagePath path) in C:\Code\NuKeeper\NuKeeper.Inspection\RepositoryInspection\ProjectFileReader.cs:line 70```

AnthonySteele:
Ack! yes, we should just skip this one ref and not crash.
"don't crash" is a good rule for almost any scenario.

skolima:
[...]

PR #315, opened 2018-07-07
Project file reader should check for no-version

AnthonySteele:
Project file reader should check for a package imported with no-version or a version range. 
These are things that happen.

e.g. `<PackageReference Include="Microsoft.AspNetCore.App" />`
and `<PackageReference Include="AWSSDK.Core" Version="3.3.*" />`
This is a very basic fix for #311 

We don't parse these cases yet. :(


AnthonySteele:
Good point. I have specifically put in a test with a version range, and done the `TryParse`.

I still want a different wording for log message for nothing vs something not parsed.

PR #316, opened 2018-07-08
Test parsing a beta package

AnthonySteele:
Test parsing a beta package
Add a test, as this is a case that wasn't covered.
And it will soon be worth looking at beta -> beta updates


AnthonySteele:
I'm not a big fan of that syntax - it seems that it's cramming unrelated concerns into the same string. But if it's a thing, we should at least have test for it. In a separate PR, though.

Issue #276, opened 2018-05-21
Global mode


AnthonySteele:
Update a particular package across all repositories on the github server.

I had this scenario presented at work:

We have an internal github at "https://github.mycompany.com".
It has multiple "[organisations](https://blog.github.com/2010-06-29-introducing-organizations/)" in github's term, used for teams or departments.
Organisations can have multiple repositories each. Most of them, but not all, contain .NET code. 
We have an internal nuget feed server at "https://packages.mycompany.com".
We have an internal nuget package, e.g. `GeneralFunctions` version `1.2.3` on this nuget feed.
A bug is discovered in `GeneralFunctions`, it poses a risk to all production systems. 
The bug is fixed in  `GeneralFunctions` version `1.2.4`, but now this fix must be rolled out across the enterprise to all projects that use this nuget package version. 
This is not easy.

Can NuKeeper automate this problem? Right now, no. The most that we can do is one org at a time. But it's in the adjacent possible - enumerate the orgs on the github server and apply NuKeeper org mode to all of them, which will in turn apply repo mode to all the repos therein. 

This could go wrong. To stop it being a "nuclear" mode, I would impose some restrictions on it, e.g.
- must be against a server other than `github.com`
- used have a `-i` package pattern to filter in only particular updates 
- A limit, max repos to process

AnthonySteele:
Done in Version 0.6.0

PR #321, opened 2018-07-09
Skipping repositories that contain no .NET projects

skolima:
This makes the Organisation mode faster, by not cloning repositories that contain no .NET projects. The check is based on either of those files being present, checked with GitHub search API:
 - packages.config
 - .csproj
 - .fsproj
 - .vbproj

### DIFF
--- a/NuKeeper.Integration.Tests/Engine/RepositoryFilterTests.cs
+++ b/NuKeeper.Integration.Tests/Engine/RepositoryFilterTests.cs
@@ -17,13 +17,13 @@ namespace NuKeeper.Integration.Tests.Engine
         {
             IRepositoryFilter subject = MakeRepositoryFilter();
 
-            var shouldSkip =
-                await subject.ShouldSkip(new RepositorySettings
+            var result =
+                await subject.ContainsDotNetProjects(new RepositorySettings
                 {
                     RepositoryName = "jquery",
                     RepositoryOwner = "jquery"
                 });
-            Assert.True(shouldSkip);
+            Assert.False(result);
         }
 
         [Test]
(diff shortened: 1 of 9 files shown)


skolima:
I've got a slight problem with this change: seems like it sometimes manages to trigger GitHub search abuse throttling...

AnthonySteele:
This could be useful, I'm working on the "global" mode for #276

NuKeeper.Integration.Tests/Engine/RepositoryFilterTests.cs
```diff
@@ -0,0 +1,49 @@
...
+            IRepositoryFilter subject = MakeRepositoryFilter();
+
+            var shouldSkip =
+                await subject.ShouldSkip(new RepositorySettings {RepositoryName = "cli", RepositoryOwner = "dotnet"});
+            Assert.False(shouldSkip);
+        }
+
+        private static RepositoryFilter MakeRepositoryFilter()
+        {
+            const string testKeyWithOnlyPublicAccess = "c13d2ce7774d39ae99ddaad46bd69c3d459b9992";
```
AnthonySteele: Ah, using this key we could also test a few more bits of the github driver.
(but not in this PR)

AnthonySteele:
It seems to me that the naming of `ShouldSkip` is not expressive and also it's a negative term - i.e. if it's true then the usefulness of that repo is false.

Can it instead be something like `ContainsDotNetProjects`, false and true in cases where `ShouldSkip` would return true and false.

NuKeeper/Engine/RepositoryFilter.cs
```diff
@@ -0,0 +1,47 @@
...
+                        $"Repository {repository.RepositoryOwner}/{repository.RepositoryName} contains no .NET code on the default branch, skipping.");
+                    return false;
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.Error("GitHub Repository search failed.", ex);
+            }
+
+            return true;
```
AnthonySteele: Is it correct to return `true` in the error case? I suppose it is, since you can't rule out .NET code.

skolima: That's the idea, yes - we will then inspect the local clone of the repo.

PR #338, opened 2018-07-17
Don't remove package before updating for dotnetcore - fixes #337

### LINKED ISSUE #337: Updating package references changes the order of packages in the csproj file
Updating package references in a `csproj` file can changes the order of the referenced projects which in turn can lead to merge conflicts and additional line noise. For example:

```
<ItemGroup>
  <PackageReference Include="A" />
  <PackageReference Include="B" />
  
  <!-- a comment -->
  <PackageReference Include="ThePackageToUpgrade" />
</ItemGroup>
```

will result in:

```
<ItemGroup>
  <PackageReference Include="A" />
  <PackageReference Include="B" />
  <PackageReference Include="ThePackageToUpgrade" />
  
  <!-- a comment -->
</ItemGroup>
```

I believe this is a side effect of removing the package reference and re-adding it (in `DotNetUpdatePackageCommand`). I think you should be able to use `add` to update if the version is specified, however I'm not an expert on doing this programatically.

Let me know if you need any more info.

skolima:
This should prevent order of packages in the file changing on updates.

### DIFF
--- a/NuKeeper.Integration.Tests/NuGet/Process/DotNetUpdatePackageCommandTests.cs
+++ b/NuKeeper.Integration.Tests/NuGet/Process/DotNetUpdatePackageCommandTests.cs
@@ -62,39 +62,44 @@ namespace NuKeeper.Integration.Tests.NuGet.Process
   <Import Project=""$(MSBuildToolsPath)\Microsoft.CSharp.targets"" />
 </Project>";
 
+        private readonly IFolder _tempFolder = UniqueTemporaryFolder();
+
         [Test]
         public async Task ShouldNotThrowOnWebProjectMixedStyleUpdates()
         {
-            await ExecuteValidUpdateTest(_testWebApiProject);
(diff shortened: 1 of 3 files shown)


AnthonySteele:
Today I'm not quite clear on how `ProjectFileOldStyle` differs from `ProjectFile` references, and why they are treated differently.

AnthonySteele:
IF I undersnad it right today, a  Net Full framework `.csproj` will have `xmlns="http://schemas.microsoft.com/developer/msbuild/2003` ([checked here](https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper.Inspection/RepositoryInspection/ProjectFileReader.cs#L60)) and may also have `PackageReference` nodes. We call this  `ProjectFileOldStyle`.

And in that case, it will need a `dotnet remove` to work with that.
But NETcore / new project style will not.

skolima:
Almost correct. VS2017 also creates those projects with the xml namespace by default, _but_ it can also work with the new project style (and the hybrid one, obviously, which is causing all the confusion here).

PR #347, opened 2018-07-22
Validate virtual

AnthonySteele:
A settings validation virtual method
The command-line framework will check as much as possible, but some things will need code
I Haven't done all the checks, there may be more to come.

### DIFF
--- a/NuKeeper/Commands/CommandBase.cs
+++ b/NuKeeper/Commands/CommandBase.cs
@@ -51,11 +51,20 @@ namespace NuKeeper.Commands
             _logger = logger;
         }
 
+
         // ReSharper disable once UnusedMember.Global
         public async Task<int> OnExecute()
         {
             _logger.SetLogLevel(Verbosity);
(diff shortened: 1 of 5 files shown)


NuKeeper/Commands/CommandBase.cs
```diff
@@ -51,11 +51,20 @@ protected CommandBase(IConfigureLogLevel logger)
...
         // ReSharper disable once UnusedMember.Global
         public async Task<int> OnExecute()
         {
             _logger.SetLogLevel(Verbosity);
             var settings = MakeSettings();
+
+            var validationResult = ValidateSettings(settings);
+            if (!validationResult.IsSuccess)
+            {
+                Console.WriteLine(validationResult.ErrorMessage);
```
skolima: I'm wondering perhaps this should write to logger Error reporting?

AnthonySteele: Well, we do have it around, so yes.

PR #324, opened 2018-07-10
Moving to CommandLineUtils

skolima:
Exploring options for #323 

Status:
- [x] `inspect` starts
- [x] `inspect` accepts options and help
- [x] logging levels can be set
- [x] `update` command works
- [x] `repository` command works
- [x] `organisation` command works
- [ ] TODO: various smaller configuration settings
- [ ] documentation update
- [ ] version output correct

This would not include configuration file support.

PR #349, opened 2018-07-24
Added the missing command line options.

skolima:
follows #324


AnthonySteele:
`ReportMode` is probably going to be reworked entirely as an output format. But we can keep it for now.

PR #350, opened 2018-07-24
read github token from env

AnthonySteele:
read github token from environment var
and refactor config validation
Populating config can't be a separate activity to validating it. 
Best example is if the URI can't be parsed,
e.g. `dotnet run repository sometoken --api foo`
then it must stop trying to populate and return a validation failure
also a couple of short command aliases: "org" and "repo"


skolima:
There's a catch with doing short command aliases like this: they're reported twice.

image

This looks a bit odd to me, TBH.

PR #351, opened 2018-07-24
Set VersionPrefix so that dotnet build picks up the versioning inform…

skolima:
…ation

Idea from https://github.com/dotnet/sdk/issues/422

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,7 +3,8 @@
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <LangVersion>latest</LangVersion>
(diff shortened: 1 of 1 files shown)


AnthonySteele:
So `VersionPrefix` is a value that the build system knows about and applies to files?

skolima:
There's a few more of those that are well-known by the sdk. https://github.com/jerriep/dotnet-outdated/blob/master/version.props this shows most of them - we could extend this to add support for build number as well.

slang25:
:+1: you could optionally remove `PackageVersion` too as it will be inferred by default, nothing wrong with being explicit though.

[Here's a great summary about how this stuff all hangs together these days](https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/)

PR #353, opened 2018-07-25
Check regex

AnthonySteele:
Validate regex parsing, and min age parsing.
So that they both exit with relevant error output if it cannot be parsed.


AnthonySteele:
I had a similar idea, a `ValidationResultWithValue<T>` where either you have a Success case with a `T` instance as the parse result, or a Failure case with a string error message. I know such types are literally 4 lines of code in Haskell, F# or Rust, but are a bit of boilerplate in C#. I tried it out but it looked a bit too "clever". I could still do it, though.

Issue #323, opened 2018-07-10
Command line options should be more alike other dotnet core tools


skolima:
> The way that dotnet does it is that a main command verb has no dash at all, e.g. `dotnet build` so `NuKeeper inspect` would work like that.
> Then the options have a short form with one dash and a long form with 2 dashes, e.g. `-v`, `--verbosity`
> And no equals signs
> e.g. `dotnet build -v q` and `dotnet build --verbosity quiet`

`CommandLineUtils` at https://github.com/natemcmaster/CommandLineUtils is a fork of the (no longer published) dotnetcore command line library and would allow us to get much closer to being consistent with other tools.

This should probably be split into multiple PRs, but published in a single new release, so as to not have multiple releases one after the other, each breaking compatibility.

Status:

 - [x] inspect starts
 - [x] inspect accepts options and help
 - [x] logging levels can be set
 - [x] update command works
 - [x] repository command works
 - [x] organisation command works
 - [x] various smaller configuration settings
 - [x] version output correct
 - [ ] review by native speakers / feedback on option names
 - [ ] documentation update
 - [ ] configuration file support (perhaps response file? that's built in)
 - [x] code cleanup (final removal of old settings handling

AnthonySteele:
Is there a document somewhere that defines how this command line syntax is supposed to look? Would help to know if we're following it.

skolima:
I don't have anything regarding spec - mostly following https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore?tabs=netcore2x to be honest.

skolima:
Main help screen 
![image](https://user-images.githubusercontent.com/95998/43189419-d439c976-8ff6-11e8-981e-a2b09794036d.png)

Inspect
![image](https://user-images.githubusercontent.com/95998/43189459-ec700604-8ff6-11e8-9170-b6190be76e40.png)

Update
![image](https://user-images.githubusercontent.com/95998/43189489-fad092ae-8ff6-11e8-9bb6-7a6b2e5fc6ec.png)

Repository
![image](https://user-images.githubusercontent.com/95998/43189520-0f556f06-8ff7-11e8-85b5-580c5eb20c8c.png)

Organisation
![image](https://user-images.githubusercontent.com/95998/43189561-20e1454c-8ff7-11e8-8301-72c427f05075.png)

Global
[...]

PR #354, opened 2018-07-28
Review Wording

AnthonySteele:
Review wording on commands and options.
for #323

### DIFF
--- a/NuKeeper/Commands/RepositoryCommand.cs
+++ b/NuKeeper/Commands/RepositoryCommand.cs
@@ -7,12 +7,12 @@ using NuKeeper.Inspection.Logging;
 
 namespace NuKeeper.Commands
 {
-    [Command(Description = "Performs version checks and generates pull requests for a single repository.")]
+    [Command(Description = "Performs version checks and generates pull requests for a single repository on GitHub.")]
     internal class RepositoryCommand : GitHubNuKeeperCommand
     {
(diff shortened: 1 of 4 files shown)


NuKeeper/Commands/RepositoryCommand.cs
```diff
@@ -7,12 +7,12 @@
...
+    [Command(Description = "Performs version checks and generates pull requests for a single repository on GitHub.")]
     internal class RepositoryCommand : GitHubNuKeeperCommand
     {
         private readonly GithubEngine _engine;
 
-        [Argument(0, Name = "GitHub repository uri", Description = "The repository to scan.")]
+        [Argument(0, Name = "GitHub repository URI", Description = "The URI of the repository to scan.")]
         // ReSharper disable once UnassignedGetOnlyAutoProperty
         // ReSharper disable once MemberCanBePrivate.Global
         protected Uri GitHubRepositoryUri { get; }
```
skolima: That one's a bit inconsistent now, as `GithubApiEndpoint` got changed from `Uri` to `string` to use `TryParse`

skolima: Separate I guess. I kind of like having CommnadLineUtils handling the whole "parse into URI" logic, but I guess having more friendly error message could be worth doing this more "manually".

PR #357, opened 2018-07-28
Update octokit by hand

AnthonySteele:
Update octokit by hand
and fix test data


AnthonySteele:
Yes, we should do that some time to avoid all this constructor breakage.
It's not quite that simple, as some fields on the the mock would still have to be set.

I want this update now though, as it has the `repo.IsArchived` flag which is a useful filter.

skolima:
Note: I had a go at using NSubstitute here, Octokit is not friendly for this. There's no interfaces used, and quite deep concrete object hierarchy with protected setters. Ugh.

PR #363, opened 2018-07-29
Octokit stubs

skolima:
This wraps the `Octokit.User` and `Octokit.Repository` into interfaces (minimal - only what we use currently is exposed). There's a few other concrete types from Octokit exposed to the code still, but those are less complex and easier to stub in test.

I'm not entirely happy with how this came out, so comments are more than welcome.


AnthonySteele:
We will have to do this some time, in a generic way some time to support gitlab, VSTS etc. since people want that. Not vital for the next release though.

Issue #365, opened 2018-07-30
Crash on Windows-1252


JornWildt:
When running NuKeeper I get this:

    Using 'inspect' for 'mode'
    Using 'https://api.github.com' for 'github_api_endpoint'
    Using '3' for 'max_pull_requests_per_repository'
    Using '10' for 'max_repositories_changed'
    Using 'verbose' for 'log_level'
    Using 'Major' for 'allowed_version_change'
    Using '7d' for 'min_package_age'
    Using 'PreferFork' for 'fork_mode'
    Using 'Off' for 'report_mode'
    Using 'nukeeper' for 'github_labels'
    Running NuKeeper in inspect mode
    Reading file C:\Projects\F2-main\NuGet.Config for package sources
    Reading file C:\Users\jw\AppData\Roaming\NuGet\NuGet.Config for package sources
    Reading file C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config for package sources
    Read [CBrain] : file://lager1/shared/fabrik/NugetRepository from file: C:\Projects\F2-main\NuGet.Config
    Read [nuget.org] : https://api.nuget.org/v3/index.json from file: C:\Users\jw\AppData\Roaming\NuGet\NuGet.Config
    Read [Microsoft Visual Studio Offline Packages] : file:///C:/Program Files (x86)/Microsoft SDKs/NuGetPackages/ from file: C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config

    Unhandled Exception: System.Xml.XmlException: System does not support 'Windows-1252' encoding. Line 1, position 31. ---> System.ArgumentException: 'Windows-1252' is not a supported encoding name. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method.
    Parameter name: name
       at System.Text.EncodingTable.internalGetCodePageFromName(String name)
       at System.Text.EncodingTable.GetCodePageFromName(String name)
       at System.Text.Encoding.GetEncoding(String name)
       at System.Xml.XmlTextReaderImpl.CheckEncoding(String newEncodingName)
       --- End of inner exception stack trace ---
       at System.Xml.XmlTextReaderImpl.Throw(Exception e)
       at System.Xml.XmlTextReaderImpl.Throw(String res, String[] args, Exception innerException)
       at System.Xml.XmlTextReaderImpl.Throw(String res, String arg, Exception innerException)
[...]

PR #366, opened 2018-07-30
Support for codepage-based encodings

skolima:


NuKeeper.Inspection.Tests/RepositoryInspection/ProjectFileReaderTests.cs
```diff
@@ -86,6 +87,21 @@ public void EmptyProjectCanBeRead()
             Assert.That(packages, Is.Empty);
         }
 
+        [Test]
+        public void UnusualEncodingProjectCanBeRead()
+        {
+            const string NoProject =
+                @"<?xml version=""1.0"" encoding=""Windows-1252""?>
```
AnthonySteele: Is this a thing that you came across?

AnthonySteele:
The exceptions look fine. 
I have never come across  `Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);` or `
System.Text.Encoding.CodePages` before.

NuKeeper.Inspection.Tests/RepositoryInspection/ProjectFileReaderTests.cs
skolima: Comes from user issue #365 - but I should have used the VS classic project in the repro here, not the core SDK one.

skolima: Checked with a dotnet core project - technically this is a valid encoding and it works with dotnetcore sdk, so I'd keep this. Test case is valid, even if unusual.

PR #382, opened 2018-08-06
Allow beta updates when the current  package version is beta

AnthonySteele:
Allow beta updates when the current  package version is beta.
This is a one line change.
Needs unit tests, an end-to-end run and a good look at the output wording.
Will it work? I don't know yet

### DIFF
--- a/NuKeeper.Integration.Tests/NuGet/Api/ApiPackageLookupTests.cs
+++ b/NuKeeper.Integration.Tests/NuGet/Api/ApiPackageLookupTests.cs
@@ -101,13 +101,11 @@ namespace NuKeeper.Integration.Tests.NuGet.Api
         }
 
         [Test]
-        public async Task BetaVersion_ShouldNotReturnBetas()
+        public async Task BetaVersion_ShouldReturnBetas()
         {
             var lookup = BuildPackageLookup();
 
-            // when we ask for updates for newtonsoft 8.0.1
-            // we know that there is a later patch (8.0.3)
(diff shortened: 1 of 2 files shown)


AnthonySteele:
I did a local test using `inspect` and `update` on a test project containing old betas of `LibGit2Sharp` and `StyleCop.Analyzers`
It appears to work just fine:

ilovebetas

NuKeeper.Integration.Tests/NuGet/Api/ApiPackageLookupTests.cs
```diff
@@ -99,6 +99,27 @@ public async Task MinorUpdateToWellKnownPackage()
             Assert.That(package.Minor.Identity.Version.Major, Is.EqualTo(8));
             Assert.That(package.Major.Identity.Version.Major, Is.GreaterThan(8));
         }
+
+        [Test]
+        public async Task BetaVersion_ShouldNotReturnBetas()
+        {
+            var lookup = BuildPackageLookup();
+
+            // when we ask for updates for newtonsoft 8.0.1
```
skolima: the comment :-)

AnthonySteele: Fixed

AnthonySteele:
It will need a sentence in the docs, yes. 

Next step, if it's wanted, would be configuration to to make this the default option, and also allow betas to be always on or always off.

Issue #383, opened 2018-08-06
When there are lots of packages to inspect a socket is used for each package


cg110:
I've been trying out nukeeper against a large set of project files, which in total have ~390 unique nugets to check.

Basically running:
nukeeper inspect -v d -c patch

I've noticed that the first updates work well (however, it looks like a fresh SSL socket is consumed for every package being checked, which is quite a lot in this case)

However, second checks seem to be very much slower, especially when the data is in cache (in fact the cache is slower than checking the web server, suggesting that the cache code in the nuget api is serialized, breaking into the process suggests that the xml parsing might be locking)

So I figure that the problem is something underlying in the nuget api usage that stops socket reuse (eg is there perhaps an object in nuget that should be Idisposable (but isn't))

I noted from debugging that PackageMetadataResource from seems to have a http source hanging off it, which is generated from the owning source repository, however, there's one sourcerepository per call.

I thought I'd try and share the source repository between all users of the same packagesource.  This appears to have fixed the socket consumption, and caching problems.

It appears that making the code share one source repository per package source fixes the caching performance.  But I'm not sure if it's a "good" fix for all code paths, as it still shows the large socket usage in the uncached case of hitting the server for data.  I suspect that perhaps the internals of the nuget library aren't making the massive concurrency good

To improve the cached case, within PackageVersionsLookup.cs, add a:
```c#
using System.Collections.Concurrent;
```
And update BuildSourceRepository as:
```c#
private ConcurrentDictionary<PackageSource, SourceRepository> _sourceCache = new ConcurrentDictionary<PackageSource, SourceRepository>();
private SourceRepository BuildSourceRepository(PackageSource packageSource)
{
    return _sourceCache.GetOrAdd(packageSource,
        key => new SourceRepository(key, Repository.Provider.GetCoreV3()));
}
```

The change pass the unit tests, but I'm not sure if it's safe for general usage or not.

I'm actually surprised that the code isn't hitting a http request limit within the .net client code, so it could be I've changed/broken something in my .net config files (but I don't recall changing anything for .net core)

skolima:
Good spot! This will need some investigation - possibly how NuGet itself is using this code could point to some mistakes we've made.

On the request limit - I'm reasonably sure dotnetcore removed those. `HttpClientHandler.MaxConnectionsPerServer` is set to `Int32.MaxValue` by default.

cg110:
[...]

PR #384, opened 2018-08-06
Cache the SourceRepository objects across parallel invocations

AnthonySteele:
Build the `SourceRepository` objects upfront
one per source, not one per request, as they are http inside
We expect that there are small number of sources, just 1 in many cases.
#383

### DIFF
--- a/NuKeeper.Inspection/NuGetApi/ConcurrentSourceRepositoryCache.cs
+++ b/NuKeeper.Inspection/NuGetApi/ConcurrentSourceRepositoryCache.cs
@@ -1,0 +1,23 @@
+using System.Collections.Concurrent;
+using NuGet.Configuration;
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+
+namespace NuKeeper.Inspection.NuGetApi
+{
+    public class ConcurrentSourceRepositoryCache
+    {
+        private readonly ConcurrentDictionary<PackageSource, SourceRepository> _packageSources
+            = new ConcurrentDictionary<PackageSource, SourceRepository>();
+
+        public SourceRepository Get(PackageSource source)
+        {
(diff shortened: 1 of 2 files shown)


AnthonySteele:
Hm, `PackageVersionsLookup.Lookup` is called from `ApiPackageLookup.FindVersionUpdate` - all one package at a time so far. 

That's called from `BulkPackageLookup.FindVersionUpdates` which processes all `PackageIdentity` in parallel using 1 `ApiPackageLookup` with 1 `PackageVersionsLookup` so we're good.

AnthonySteele:
Yeah, after thinking about it some more, @cg110 had the code right:
`PackageVersionsLookup.Lookup` is going to get called a lot of times, and sources should be shared between these invocations. 
And they happen in parallel, so `ConcurrentDictionary` is the right way to cache them in the `PackageVersionsLookup`.

NuKeeper.Inspection/NuGetApi/ConcurrentSourceRepositoryCache.cs
```diff
@@ -0,0 +1,23 @@
...
+            = new ConcurrentDictionary<PackageSource, SourceRepository>();
+
+        public SourceRepository Get(PackageSource source)
+        {
+            return _packageSources.GetOrAdd(source, CreateSourceRepository);
+        }
+
+        private static SourceRepository CreateSourceRepository(PackageSource packageSource)
+        {
+            return new SourceRepository(packageSource, Repository.Provider.GetCoreV3());
```
AnthonySteele: This line is equivalent to before, but could be changed in a later PR.
The issue is that it always adds the source as a Nuget protocol V3 source.
But you can use the property `int PackageSource.ProtocolVersion` to check.
I don't know what you would do if it's V2 or something else though.

PR #387, opened 2018-08-08
Better detailed logging in DotNet and nuget commands

AnthonySteele:
In Detailed verbosity, Log the complete text of the actual commands sent to `dotnet` or `nuget` in a standard format.


skolima:
Uhm, instead of having the logging on usage, shouldn't we rather bring back the logging of the full command in the ProcessRunner (in Diagnostic level, obviously)? We had this before the refactor, but forgot to re-add it.

AnthonySteele:
Hm, yes it's got to the point where it's effectively that anyway.

PR #394, opened 2018-08-16
Convert a Creator to a parameter

AnthonySteele:
This _might_ be the right way to go. I tried removing the `ICreate` infrastructure from `UpdateSelection` by passing the settings in via the entry point method not the constructor. The rest is the ripple effect. Two creators were removed.


AnthonySteele:
I agree, `ICreate` is a reasonable short-term plan to use Commands. But slowly refactoring them out is also good.

PR #395, opened 2018-08-16
Use FilterSettings more

AnthonySteele:
More refactoring around settings.

`FilterSettings` needs to exist, it and the code that consumes it is in `NuKeeper.Update` and cannot reference `NuKeeper.Configuration` namespace for `UserSettings`. 

However the same fields don't need to be there on `UserSettings`, the `SettingsContainer` can instead also own a `FilterSettings` and therefore there is no need for translation in `FilterSettingsCreator`.


NuKeeper/Configuration/UserSettings.cs
```diff
@@ -10,16 +10,8 @@ public class UserSettings
...
-
-        public Regex PackageExcludes { get; set; }
-
-        public int MaxPackageUpdates { get; set; }
-
         public int MaxRepositoriesChanged { get; set; }
 
-        public TimeSpan MinimumPackageAge { get; set; }
-
         public VersionChange AllowedChange { get; set; }
```
skolima: Shouldn't `AllowedChange` also belong on the `PackageFilter`?

AnthonySteele: `FilterSettings` is currently all the fields used in `UpdateSelection. `AllowedChange` is used elsewhere. It's purely locality of code rather than being conceptually related.

Issue #386, opened 2018-08-07
Unable to use NuKeeper


JBoothUA:
I seem to be running into an issue I was hoping someone could point me in the right direction?
This basically happens on any update command:

---

D:\Repos\Galileo\CommonSite\Source\CommonSiteService [master ≡ +0 ~1 -0 !]> NuKeeper update --include Mvc
Found 46 packages in use, 46 distinct, in 1 projects.
[...]

PR #404, opened 2018-08-21
Fixes local source path escaping

skolima:


skolima:
Related to #386

AnthonySteele:
`ArgumentEscaper.EscapeAndConcatenate` looks useful, but what is actually going wrong with adding `"` quotes? Are some of the args already `IsSurroundedWithQuotes` ?

skolima:
I was adding a) quotes where they weren't required (see the updated test - there's no need to quote the first source) and b) not escaping the `/` where it was required (thus the source `C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\` was failing. CMD.exe is odd in that `\` only needs to be escaped sometimes. Fun!

AnthonySteele:
Ah, so _a_ was not fatal but _b_ was. Well spotted! and good that someone already has code to do the right thing with it!

Issue #339, opened 2018-07-17
Local config.json


shatl:
Hi,
Is this possible to have local `config.json` in the directory I'm running nukeeper? 

Also, please rename config.json to something more specific (e.g. `nukeeper.json` or `nukeeper.config.json`).

P.S. Thanks for a great work done on nukeeper!

AnthonySteele:
Hi

We're currently working on revamping settings, and the config file. These are good ideas, thanks.

AnthonySteele:
Command-line args are all different in version `0.6`. Config file support might come back in some form, but probably using names identical to the "long name" of command line args. Which options would you use from file?

shatl:
Hi Anthony,

* package sources;
* age;
* include / exclude masks;
* verbosity;
* label;
* api endpoint configuration;

AnthonySteele:
I am having a look at reading settings from a `nukeeper.config.json` file in the current directory.

I am starting with `Include` and `Exclude` regexs, and will also look at max changes etc.

I do not see that putting pakage sources in this file is  necessary: by default, NuKeeper will read the `NuGet.config` file of the target code and take settings from there, as the build also does.

AnthonySteele:
All the fields of this DTO are now settings that can come from the file `nukeeper.config.json` in release 0.7: 

https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper/Configuration/FileSettings.cs

And will work as the same-named commandline option.

This covers all the options listed above, with the exception of package sources. These should come from `NuGet.config` instead.

Better docs will follow in the next few days.

shatl:
Thanks a lot, works for me

PR #414, opened 2018-09-02
File settings

AnthonySteele:
Config file for persisting commandline settings
#339

### DIFF
--- a/NuKeeper.Tests/Configuration/FileSettingsReaderTests.cs
+++ b/NuKeeper.Tests/Configuration/FileSettingsReaderTests.cs
@@ -1,0 +1,133 @@
+using System.IO;
+using NSubstitute;
+using NuKeeper.Configuration;
+using NuKeeper.Inspection.Files;
+using NuKeeper.Inspection.Logging;
+using NUnit.Framework;
+
+namespace NuKeeper.Tests.Configuration
+{
+    [TestFixture]
+    public class FileSettingsReaderTests
+    {
+        [Test]
+        public void MissingFileReturnsNoSettings()
+        {
+            var folder = TemporaryFolder();
+
+            var fsr = new FileSettingsReader(Substitute.For<INuKeeperLogger>());
+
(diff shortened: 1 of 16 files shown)


AnthonySteele:
Consider this a proof of concept. So far the file format is simple, json file containing key-value pairs. The keys should all be the same as the long form of a command line switch. Capitalisation of keys should not matter.

What I want to sort out soon is the rest of it - the file name, the file path(s), how the code parses it. The current code is OK but not great.

It is possible to have a commandline arg to use a named file, e.g. `--settingsfile c:\some.file.json` but I would leave that to a later PR.

AnthonySteele:
Work has got further, there are now 5 string values in the config file.
And a test on reading it.

AnthonySteele:
I have had a play with doing this via the .NETCore settings api instead of Newtonsoft and file reading.

It's a simple replacement, entirely inside `FileSettingsReader`. The code looks like this:

```csharp
        private FileSettings ReadFile(string folder, string fileName)
        {
            try
            {
                var builder = new ConfigurationBuilder()
                    .SetBasePath(folder)
                    .AddJsonFile(fileName);

                var settings = builder.Build();
                return settings.Get<FileSettings>();
            }
            catch (Exception ex)
            {
                _logger.Error($"Cannot read setings file at {folder} {fileName}", ex);
                return FileSettings.Empty();
            }
        }
```

Pro: this opens the door to easily reading from more than one file, env vars etc using the fallback mechanisms in the config api.

Con:  Needs _three_ new packages added to the project: `Microsoft.Extensions.Configuration`, `"Microsoft.Extensions.Configuration.Binder` and `Microsoft.Extensions.Configuration.FileExtensions`

Deciding factor: Not going to change over to settings api now as don't need it to only read 1 file. 

Doing this later if necessary seems not hard -  it would be a change in 1 class, `FileSettingsReader`, and will not ripple through the codebase.

Issue #128, opened 2017-08-30
Batching


AnthonySteele:
Some packages are best updated together, e.g. `NLog` and `NLog.Schema`, or all the `AWSSDK.*` packages.
Roll them into 1 PR.

skolima:
Note: `Microsoft.AspNetCore.All` is  another major example here, if the individual packages are updated but not the "meta-package", the build will fail.

daniel-munch-cko:
Just gave NuKeeper a try and it's really rad. Thanks a lot! 

Are there plans or ideas for the batching to be implemented soon? I think it would be a real useful addition. If nobody's currently working I'd be happy to give it a look.

How would you imagine such a feature? I've been thinking about either
- making it configurable, i.e. explicitly say that `AWSSDK.*` or `Microsoft.Extensions.*` should be batched
- do it automagically, i.e. find the LCD of all packages being updated from the same version towards the same version.

Thoughts?

skolima:
Automatic detection would have to be quite clever to deal with all the cases we've seen so far - perhaps some default settings in the configuration file for the well known package sets, along with allowing the user to alter the setting to deal with internal packages that have to go in step, would be a better approach?

AnthonySteele:
Hi.

We have no WIP on batching right now.

My preference is to use automation for this as much as possible. My thinking is that NuKeeper *is* automation, because .net devs don't manually update very well, because a tool to proactively check for and apply updates raises the game.

I'm not quite sure how, but I do know that the package info that the Nuget API returns includes dependencies. A first step would be in the ordering - if there are updates to both `AWSSDK.Core` and `AWSSDK.Lambda`, and Lambda depends on Core, then the update to Core should be sorted above the update to Lambda.

skolima:
I've just hit another instance where this would be very useful - and where Visual Studio updater gets it wrong as well, currently. `Microsoft.AspNetCore.App` 2.1.1 rolled out recently, and with it a large group of other updates turns out e.g. those need to be updated in sync or the project won't build any more:
`Microsoft.AspNetCore.App`
`Microsoft.VisualStudio.Web.CodeGeneration.Design`
`Microsoft.AspNetCore.TestHost`
`Microsoft.Extensions.Configuration`

What makes it trickier is that, like `System.IO.Abstractions` and `System.IO.Abstractions.TestingHelpers`, some of those packages would be installed into a "main" projects and some into a test project - and they need to match.

[...]

PR #421, opened 2018-09-10
Add ability to consolidate updates into single pr

cookie1981:
We've added a new config option `--consolidate` (defaults to false) which will allow you to push all your updates into a single PR when multiple packages have been updated.

Works for `repo`, `org`, `global` commands.


cookie1981:
@AnthonySteele will resolve conflicts asap - just in a meeting

AnthonySteele:
So this seems to be a manual take on "batching" or #128
If I do a run with `--include ^AWSSDK. --consolidate` then I should get one PR from that run for all the updated to AWS SDK packages (there are  numerous ones, all released together).
Downside that you won't get anything else from that run.

NuKeeper/Commands/GitHubNuKeeperCommand.cs
```diff
@@ -1,136 +1,142 @@
-using System;
```
skolima: There's EOL changes here - is that a mismatch in your settings or is the file in GitHub already mangled? If that the latter, I'd rather do a EOL fix commit separately first.

cookie1981:
I hadn't seen the open issue, but that yes, that is one use case where this feature would be useful.
The other being that all updates could be rolled up into one PR if you wanted, although personally I'd prefer to have related things in one PR.

I actually don't see the behaviour above as a 'downside', more a desired behaviour: I'm combining a number of switches together to 'focus' NuKeeper on a specific areas.

We are running this in automated pipelines.
One pipeline updates dependencies (excluding Microsoft.AspNetCore), creating one PR per dependency.
One pipeline updates some internal libraries we build in-house, but we want to package them all up together so that it's atomic...otherwise we'd end up with 2 or 3 PRs, both failing due to the absence of the others.

NuKeeper/Commands/GitHubNuKeeperCommand.cs
skolima: I think that's on your side - added #423 which should prevent line endings issues in the future (unless, of course, someone uses a single checkout both with Linux and Windows tooling...)

acraven: It appears that #420 swapped from LF to CRLF and as we merged conflicts we unknowingly switched it back to LF. I've added a comment in #423 as such.

cookie1981:
I've re-read #128 and can see where you are coming from with the automation comment...
perhaps this could act as a first step in moving towards that, then iterate to move towards the bigger vision.

NuKeeper.Tests/Engine/RepositoryUpdaterTests.cs
```diff
@@ -28,12 +33,11 @@ public class RepositoryUpdaterTests
         [Test]
         public async Task WhenThereAreNoUpdates_CountIsZero()
         {
-            var packageUpdater = Substitute.For<IPackageUpdater>();
             var updateSelection = Substitute.For<IPackageUpdateSelection>();
             UpdateSelectionAll(updateSelection);
 
-            var repoUpdater = MakeRepositoryUpdater(
-                packageUpdater, updateSelection,
+            var (repoUpdater, packageUpdater) = MakeRepositoryUpdater(
```
AnthonySteele: New syntax?

NuKeeper/Creators/PackageUpdaterCreator.cs
```diff
@@ -21,6 +21,11 @@ public PackageUpdaterCreator(INuKeeperLogger logger, ICreate<IGitHub> githubCrea
 
         public IPackageUpdater Create(SettingsContainer settings)
         {
+            if (settings.UserSettings.ConsolidateUpdatesInSinglePullRequest)
+            {
+                return new ConsolidatingPackageUpdater(_githubCreator.Create(settings), _updateRunner, _logger);
+            }
+
```
AnthonySteele: The one thing that stands out is that we want to move away from using the `ICreate` interfaces where possible. So adding code into them will work against this. Possibly we want to instead lean on the IoC container and give these 2 classes different interface types?

NuKeeper.Tests/Engine/RepositoryUpdaterTests.cs
cookie1981: Tuple deconstruction. Added in 7.Something

AnthonySteele:
So I like this PR (in some form) for the simple reason that I would use it.  We use AWS a lot at work, and I would use the `--include ^AWSSDK. --consolidate` run.

cookie1981:
ok. So, what do we need to do to get this merged in then? 
I assume that you want us to refactor inline with the above comment about the creator? And that's what you mean about "in some form"

AnthonySteele:
Yes indeed, refactor in line with that comment.
Sorry that I haven't had a lot of time to review the code in detail. I planned to take time this week to advance other aspects parts of the code, which is happening. So this came from left field.  Nevertheless, I think it is a likely next step in slow progress towards #128

AnthonySteele:
The thing that I think is against my design ideas is the 2 implementations of `IPackageUpdater`

Batching is a good idea, but has been shelved so far, because there have been many other things to do, and it it is not easy to do batching correctly and/or automatically. So we have to get there in small steps.

IMHO: There should be one `IPackageUpdater`, and it is should be a collection of `PackageUpdateSet` instead of 1. Often this list will contain 1 `PackageUpdateSet`. It will apply updates locally and then make the appropriate Pull Request.

The task of "grouping" or "consolidating" package update sets can be made a separate thing, and the MVP for this is your proposal, either 1 `PackageUpdateSet` per PR, or 1 PR containing all `PackageUpdateSet` together.

Something like:

```csharp
public static class PackageUpdateConsolidator
{
	public static IReadOnlyCollection<IReadOnlyCollection<PackageUpdateSet>> Consolidate(
		IReadOnlyCollection<PackageUpdateSet> updates, bool consolidate)
	{
		if (consolidate)
		{
			return new List<IReadOnlyCollection<PackageUpdateSet>> { updates };
		}

		return updates.Select(u => new List<PackageUpdateSet> {u}).ToList();
	}
}
```

This way, if someone makes a "clever" consolidator then the `IPackageUpdater` does not need to change, it is just adding a new grouping strategy to the above.

Does this make sense?

PR #423, opened 2018-09-11
Set the equivalent of `core.autocrlf=true` in .gitattributes

skolima:
Follows https://github.com/dotnet/core/blob/master/.gitattributes


acraven:
Hi, can I just confirm what the line endings should be please? Looking at the editor config file it suggests CRLF. I'm using Rider on Mac and it's showing LF for all files (I've checked) except GitHubNuKeeperCommand (as per comments in #421). VsCode also shows the same.

skolima:
Oh. Yeah, that's an error in the `.editorconfig` - we've been developing on Windows before. I'll add another change to this thing then, give me a sec :-)

BTW, related : https://github.com/editorconfig/editorconfig/issues/226 (rejected proposal)

acraven:
It looks to me like #420 changed the line endings in GitHubNuKeeperCommand. Can I suggest that this PR includes a fix to revert this file to LF and also update the .editorconfig accordingly? Assuming that is, that LF is the desired ending.

acraven:
Mmmm, this changes looks like they've changed one way or another...
https://github.com/NuKeeperDotNet/NuKeeper/pull/420/commits/3db18902a663c6c8df7d8dc2809b27f760542113

I have a video of what I'm seeing in rider suggesting the above changes the ending from LF to CRLF...
https://www.useloom.com/share/fce10b3b5aec402199a65cc40ddc48b7
(sorry, would have been better with sound, but keep an eye on the ending in the status bar)

skolima:
@acraven could you check _this_ branch? What's Rider showing as line endings?

skolima:
Good old `dos2unix` to the rescue. This should now work as expected.

AnthonySteele:
I have a new laptop this week, "fighting with git line endings" sounds about right.

Issue #439, opened 2018-09-19
Throws System.ArgumentException while reading a local NuGet.config file


msschl:
#### Steps to reproduce
- Create a new project
- Place a NuGet.config file in the project root directory with the follwing content:
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <!-- Restore sources should be defined in build/sources.props. -->
  </packageSources>
</configuration>
```
- Run the command `nukeeper inspect -v d`

#### Produces
```shell
Reading file F:\Programming\c#\Geocoding\NuGet.Config for package sources
Reading file C:\Users\XXX\AppData\Roaming\NuGet\NuGet.Config for package sources
Reading file C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config for package sources

Unhandled Exception: System.ArgumentException: sources
   at NuKeeper.Inspection.Sources.NuGetSources..ctor(IEnumerable`1 sources) in C:\Code\NuKeeper\NuKeeper.Inspection\Sources\NuGetSources.cs:line 38
   at NuKeeper.Inspection.Sources.NuGetConfigFileReader.ReadFromFile(IReadOnlyCollection`1 sources) in C:\Code\NuKeeper\NuKeeper.Inspection\Sources\NuGetConfigFileReader.cs:line 41
   at NuKeeper.Inspection.Sources.NuGetConfigFileReader.ReadNugetSources(IFolder workingFolder) in C:\Code\NuKeeper\NuKeeper.Inspection\Sources\NuGetConfigFileReader.cs:line 30
   at NuKeeper.Inspection.Sources.NuGetSourcesReader.Read(IFolder workingFolder, NuGetSources overrideValues) in C:\Code\NuKeeper\NuKeeper.Inspection\Sources\NuGetSourcesReader.cs:line 26
   at NuKeeper.Local.LocalEngine.Run(SettingsContainer settings, Boolean write) in C:\Code\NuKeeper\NuKeeper\Local\LocalEngine.cs:line 43
   at NuKeeper.Commands.InspectCommand.Run(SettingsContainer settings) in C:\Code\NuKeeper\NuKeeper\Commands\InspectCommand.cs:line 33
   at NuKeeper.Commands.CommandBase.OnExecute() in C:\Code\NuKeeper\NuKeeper\Commands\CommandBase.cs:line 69
[...]

PR #440, opened 2018-09-19
Sources reliability

AnthonySteele:
Make package sources more reliable
fix for #439


skolima:
I've got... mixed feelings here. Someone explicitly configured their project to use no update sources. Shouldn't NuKeeper rather just report this and exit?

AnthonySteele:
It probably won't fix their case - NuKeeper will eventually run standard `dotnet add` commands, and as far as I have seen, this will only update the `.csproj` files, not the custom `build.props` files etc. But it still should not crash.

Issue #331, opened 2018-07-12
Not all "inspect" output is logging


AnthonySteele:
I'd like a bit of a rethink with the output of the inspect command, and logging, and the csv report which should be folded in there. 

Currently all output goes via the logger to the console; except for reports, which are csv to file. I'm thinking that the inspect command is run in order to get this output, and it's something different to the debug logging. You will want to split into two outputs that can be managed independantly. 

The non-logging output might have a choice of output formats (e.g. readable text, csv report, json, just the libyear score), and a choice of output destinations (stdout, named file). or possibly configure so that the inspect output goes to stdout and the logging to file.

AnthonySteele:
So, when NuKeeper is run against github repo(s) the output is useful only for debugging.
When a local inspect is run, the output is everything.
You might want something readable, or  a `.csv` format report, or just the libyears score.
You might want that to go to file, or you might want that to go to console and the debug logging to go to file.

AnthonySteele:
Mostly done in `v.0.9.0`

PR #442, opened 2018-09-21
Loggers

AnthonySteele:
Make logging more flexible. `--logfile somefile.log` will append output to that file.

WIP towards #331 , and generally making the local inspect output nicer and more flexible, and reworking the report output


NuKeeper.Inspection/Logging/IConfigureLogger.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    public enum LogDestination
+    {
+        Console,
+        File
+    }
+
+    public interface IConfigureLogger
+    {
+        void Initialise(LogLevel logLevel, string filePath);
```
skolima: Uhm, that file is a bit odd. Enum isn't related to the interface, and the interface is mixing log levels and file logging.

NuKeeper.Inspection/Logging/IInternalLogger.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+
+namespace NuKeeper.Inspection.Logging
+{
+    public interface IInternalLogger
+    {
+        void Error(string message, Exception ex = null);
```
skolima: Personal bias against optional parameters, but: this makes it way easier to forget to include the actual exception when logging. I'd vote for explicit `null` in those cases.

AnthonySteele: Yep, the default is no longer needed on the internal logger.

skolima:
`--logfile` seems to be the standard switch https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2017

PR #446, opened 2018-09-25
IsPackable is not needed

AnthonySteele:
"IsPackable: A Boolean value that specifies whether the project can be packed. The default value is true."
https://docs.microsoft.com/en-us/dotnet/core/tools/csproj


slang25:
On the same theme, `RootNamespace` and `AssemblyName` default to the project filename so could also be removed.

PR #450, opened 2018-09-27
Consolidate

AnthonySteele:
WIP based on #421

>We've added a new config option --consolidate (defaults to false) which will allow you to push all your updates into a single PR when multiple packages have been updated.

> Works for all github commands: `repo`, `org`, `global`.


AnthonySteele:
There are opportunities for improving wording or refactoring tests to remove duplicate data creation helper methods; but those are best left to later PRs.

cookie1981:
Oh nice one - I've been on holiday and was just looking at picking this up again. Thanks (& sorry for the radio silence)

Issue #453, opened 2018-10-01
Regression: command line fork mode ignored in 8.0.0


skolima:
Command line:
`NuKeeper org MyOrg %GitHubAccessToken% --maxpr=50 --maxrepo=500 --source=https://api.nuget.org/v3/index.json --source=%internal-feed% -v:d --fork=SingleRepositoryOnly --label=nukeeper --label=review-requested`

0.7.2 results in:
`FindPushFork. Fork Mode is SingleRepositoryOnly`
0.8.0 results in:
`FindPushFork. Fork Mode is PreferFork`

pgrm:
having same issu

AnthonySteele:
Does it do the same with spaces not equal signs, e.g. `--fork SingleRepositoryOnly` ?

AnthonySteele:
Cause seems to be that the commandline is being read just fine, but the `ForkFinder` is being created before that.

AnthonySteele:
Will be fixed in `0.8.1`

AnthonySteele:
Version `0.8.1` should be visible shortly.

PR #454, opened 2018-10-01
Fix fork mode

AnthonySteele:
Fix for #453

The commandline parsing was fine, the issue was in IoC ordering changes.

It is no longer correct for `ForkFinder` to get settings via the constructor and read the ForkMode off of it, as this happened before the setting were initialised.

Refactored to a method with another param but less object state.  Semi-functional style.


AnthonySteele:
It's _possible_ that similar issues could occur elsewhere, but what would be needed would be to inject settings via constructor and then store not a ref to the settings, but a value out of it. Neither are recommended any more.

PR #459, opened 2018-10-02
Add support for version Release option

### LINKED ISSUE #457: Include preview version/changes in filter
Hi,

How can we include preview versions in the filter?
I have this situation:
```
Found 1 possible updates
nanoFramework.CoreLibrary from 1.0.1-preview094 to 1.0.1-preview103 in packages.config
```

But I can't persuade the update to happen, it's always being filtered out...

josesimoes:
- Add Release to VersionChange enum.
- Update code to process nuget version _Release_ (like in X.Y.Z-alpha) as a valid update criteria.
- Add/update tests accordingly.

- Fix #457.

⚠️  couldn't edit the Wiki, so this will require updating the description for the "change" option.

Signed-off-by: José Simões <jose.simoes@eclo.solutions>

### DIFF
--- a/NuKeeper.Inspection/NuGetApi/VersionChanges.cs
+++ b/NuKeeper.Inspection/NuGetApi/VersionChanges.cs
@@ -19,7 +19,8 @@ namespace NuKeeper.Inspection.NuGetApi
             var major = FirstMatch(orderedCandidates, current, VersionChange.Major);
             var minor = FirstMatch(orderedCandidates, current, VersionChange.Minor);
             var patch = FirstMatch(orderedCandidates, current, VersionChange.Patch);
-            return new PackageLookupResult(allowedChange, major, minor, patch);
+            var release = FirstMatch(orderedCandidates, current, VersionChange.Release);
+            return new PackageLookupResult(allowedChange, major, minor, patch, release);
         }
 
         private static PackageSearchMedatadata FirstMatch(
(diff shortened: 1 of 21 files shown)


NuKeeper.Inspection/NuGetApi/VersionChanges.cs
```diff
@@ -43,6 +44,23 @@ private static bool Filter(NuGetVersion v1, NuGetVersion v2, VersionChange allow
...
+                        if (string.IsNullOrEmpty(v2.Release))
+                        {
+                            return (v1.Major == v2.Major) && (v1.Minor == v2.Minor);
+                        }
+                        else
+                        {
+                            return (v1.Major == v2.Major) && (v1.Minor == v2.Minor) && (v1.Patch == v2.Patch);
+                        }
+                    }
+
```
AnthonySteele: I think this is the key code.  Can you explain why the `Patch` is only checked if both have `Release` tags? Versions that differ only in `Release` would have same Major, Minor and Patch parts?

Also, I would start it with 
``` 
if (string.IsNullOrEmpty(v1.Release)) || string.IsNullOrEmpty(v2.Release))
{
 return ...
}
```

josesimoes: Looking at the calling code, this is comparing the `current` version with the available versions collection.


1. If the `current` doesn't have a release tag, it won't matter if the comparing one has one or not, as long as it's higher or the same, it's good to go. Doesn't matter the release here, because, even on the most de-favourable case 1.2.3 against 1.2.3-preview002, the 1st will always win because the second is a preview of the first.

2. If the comparing one doesn't have a release tag, the comparison should be the same as with Patch, because it won't matter.


3. If the comparing one _has_ a release tag (like the current, than we need to compare all three, up to the release one.

AnthonySteele: In the existing code , If the allowedChange is `Minor`, it won't match `1.2.0` to `2.3.0` since the Major is different and only Minor or below can differ. If the allowedChange is `Patch`, it won't match `1.2.0` to `1.3.0` since the Minor is different and only Patch can differ.

You'd expect based on that for allowedChange of `Release` then the Major, Minor and Patch should all match; but it's not the same since release number is optional. So we're saying that `Release` should work as `allowedChange` of `Patch`, unless it's comparing both current and available that both have non-empty release values?

AnthonySteele:
An allowed change of "major" should pick up everything, and a current package with a release tag should pick up updates with same. So it might be filtered out later instead?

josesimoes:
You are correct... somehow I screw up when testing this and the "major" option wasn't updating the project as intended. I've reverted all this and using "major" it does update anything just like you described.

Apologies for taking your time with all this. 😞

PR #479, opened 2018-10-18
Multipackage wordings

AnthonySteele:
Update to mutlipackage commit wording, 
 * Header says how many packages _and projects_ are affected.
 * The long details is under a collapse / expand element, [similar to this](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab).
 * a unique branch name for consolidated update PR, containing 2 or more package updates.

A unique branch name is needed because you can't use the same branch name each time, it will fail the second time because the branch already exists. Unless you're diligent about removing branches.

When there are 2 more more updates in the PR, the branch name has a  MD5 Hash on the end.
The input to the hash is the list of target package names and versions.

We want the same set of updates to generate the same branch name, but we also don't want the branch name to be arbitrarily long. So we hash the names and versions into a fixed-length string.
 
Yes, we know that the MD5 hash fn isn't secure, but we don't see a security implication for using this to try to make branch names unique. It's hard to give it bad input data, as this is read from the nuget api. Worst case is that the branch name isn't unique and so NuKeeper can't make the PR. 

Prompted by the good and bad parts of this PR: https://github.com/justeat/AwsWatchman/pull/214

### DIFF
--- a/NuKeeper.Tests/Engine/BranchNamerTests.cs
+++ b/NuKeeper.Tests/Engine/BranchNamerTests.cs
@@ -49,6 +49,28 @@ namespace NuKeeper.Tests.Engine
         }
 
         [Test]
+        public void EquivalentInputs_HaveSameHash()
+        {
+            var packages1 = new List<PackageUpdateSet>
+            {
+                PackageUpdates.MakeUpdateSet("SomePackage", "2.3.4"),
+                PackageUpdates.MakeUpdateSet("OtherPackage", "2.3.4")
+            };
+
+            var packages2 = new List<PackageUpdateSet>
+            {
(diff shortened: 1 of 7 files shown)


skolima:
Wow that PR had a wall of text!

Caveat to the branch name generation: this won't fail branch creation now, but, on the other hand, the mechanism of "I've already generated a PR for this, because a branch exists" won't work either. Perhaps instead of a timestamp, a hash all the included packages and versions could be used? Then if any of them change in a following run, a new branch would be generated, but identical runs would not create new branches.

AnthonySteele:
Yep, the "wall of text"  needs to go behind a [Click to expand thingy](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab)

Issue #336, opened 2018-07-15
Add support for VSTS


OneCyrus:
It would be great to have VSTS support. This should be pretty similar to implement like GitHub. 

There's a GitHttpClient to do this: 
https://msdn.microsoft.com/library/microsoft.teamfoundation.sourcecontrol.webapi.githttpclient.aspx

it's published as a nuget package: 
https://www.nuget.org/packages/Microsoft.TeamFoundationServer.Client/16.136.1-preview

AnthonySteele:
Related to #327

As either of these means making the remote source control system "plugable"

JBoothUA:
this would be wonderful!

phatcher:
@AnthonySteele I'm going to have a go at this as it would be useful for a couple of my projects

AnthonySteele:
My thinking is that it could proceed along these lines:

- pull the github driver code out in `OctokitClient ` into a separate project that references `Octokit`. leave the `IGitHub` interface that it implements where it is
- Where it returns github types, e.g. `Account` we will need wrapper types that have the few fields that we need.
- rename the `IGitHub` interface something more neutral.
- provide a second implementation of this interface, in a project that references the VSTS driver.

phatcher:
That's the sort of thing I was thinking of.  

VSTS has a REST API and an [associated client](https://www.nuget.org/packages/Microsoft.VisualStudio.Services.Client/), though when I tried to install it, the .NET Standard dependencies seem a bit borked, so I might just do a "pure" REST client using HttpClient - was trying to avoid this due to authentication etc.

So we put new wrapper interfaces/classes in NuKeeper and then proxy them for each provider type

AnthonySteele:
great. The only thing that i would add is, if we know where we're going, small incremental PRs work best. There will be other dev work ongoing during this.

phatcher:
I'll work independently initially i.e. get a working VSTS client before I attempt to do the change that affects any of the existing code.

That way I'll know exactly what needs to be proxied and whether there is an impedance mis-match between the models in NuKeeper, GitHub and VSTS

MrGlenThomas:
@phatcher how's it coming along? I work with VSTS and git and have written a tool for creating and managing git branches so have some code for that.

shep1987:
Really is a proof of concept but if anyone else it looking at this, this may help:

https://gist.github.com/shep1987/2647d0b6e1264c2f102c79d65edfa2db

[...]

PR #480, opened 2018-10-20
[WIP] Abstract workings so that we can add other client(such as AzureDevOps)

MarcBruins:
First of all, i love the idea of NuKeeper and i feel very excited about it! This is my attempt to make the solution more generic so that we can easily add new client such as AzureDevOps or bitbucket.

The proposed solution structure will look like this:

- NuKeeper(Contains the command and the DI for registering the different clients)
- NuKeeper.Abstract (Contains the abstractionlayer for the Engine, the Configuration and some other shared code)
- NuKeeper.Github(Contains github implementation)
- NuKeeper.AzureDevOps(Contains AzureDevOps implementation)
- And the rest of the already existing parts

To let code function the same way for differenct clients i've extracted interfaces and used Githubs Octokit as a reference. I've used sharedcode where i can and made an implementation for Github specifically. So far everything builds but it doesn't work yet. I'm still working on getting the Github functionality back and would like to advance from there on to add AzureDevops.

The progress of this PR is:
- [x] Successful build with new project styles and abstraction layer
- [ ] Verify that all the same functionality that the package now has still works
- [ ] Update tests

This is WIP, please let me know what you think!


NuKeeper.Abstract/Engine/Packages/ExistingBranchFilter.cs
```diff
@@ -1,29 +1,28 @@
...
 
         public async Task<bool> CanMakeBranchFor(PackageUpdateSet packageUpdateSet,
-            ForkData pushFork)
+            IForkData pushFork)
         {
             try
             {
                 var branchName = BranchNamer.MakeName(packageUpdateSet);
-                var githubBranch = await _gitHub.GetRepositoryBranch(pushFork.Owner, pushFork.Name, branchName);
+                var githubBranch = await _client.GetRepositoryBranch(pushFork.Owner, pushFork.Name, branchName).ConfigureAwait(false);
```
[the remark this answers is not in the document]
MarcBruins: Good point, i didn't know that. I'll change it back.

AnthonySteele:
This is related to #336 

I haven't had time to look at this yet, I am away from home this weekend. Will do so when  can.
But it is in general a good idea to go in this direction of abstracting out github and generalising to other systems, hopefully in small steps.

NuKeeper.Abstract/NuKeeper.Abstract.csproj
```diff
@@ -0,0 +1,18 @@
...
+        <TargetFramework>netstandard2.0</TargetFramework>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\NuKeeper.Inspection\NuKeeper.Inspection.csproj" />
+      <ProjectReference Include="..\NuKeeper.Update\NuKeeper.Update.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <Reference Include="LibGit2Sharp, Version=0.26.0.0, Culture=neutral, PublicKeyToken=7cbde695407f0333">
```
AnthonySteele: This should be a `PackageReference` not a relative pathed dll reference. Issue with tooling?

NuKeeper.Abstract/NuKeeper.Abstract.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
+    </PropertyGroup>
+
```
AnthonySteele: New libraries should also contain:
````
  <PropertyGroup>
    <CodeAnalysisRuleSet>..\CodeAnalysisRules.ruleset</CodeAnalysisRuleSet>
  </PropertyGroup>
````

AnthonySteele:
So, a general design comment around things in `NuKeeper.Abstract` such as `interface IPullRequest` and `class GithubBranch : Branch, IBranch`

We're going to have to get this working well, and the plan (here)[https://github.com/NuKeeperDotNet/NuKeeper/issues/336#issuecomment-416992498] was to instead use Nukeeper's own types that "have only the few fields that we need". 

We are aware that the Octokit types are not ideal to use inside NuKeeper - the biggest symptom of that is tests where creating an Ocotkit object requires over 30 args to the constructor (!).  At some stage this might be changed to use github's graphQL API, and retrieve "only the few fields that we need". 

But it seems odd to me to use Interfaces and inheritance to solve this issue. Ideally I'd like to see all backends work on a common set of internal DTOs, that don't need interfaces or base classes. This does depend of the functionality being more or less equivalent so that common data types make sense.

AnthonySteele:
Big PRs with lots of changes and also merge conflicts are my least favourite kind :(
There must be ways to slice this up into things that can be merged today, e.g. introduce just the "abstractions" project. Or just the "github" project.

I am happy with a PR that doesn't make a visible difference in itself; I think we have reasonable idea of where you are going.

MarcBruins:
I would like to close this so that we can keep PR's smaller as @AnthonySteele mentions. That doesn't mean that i'm throwing away the progress that's been made. This PR can be kept as a reference towards a more abstract solution which enables more clients. I'll raise a issue with steps that i see so we can make a list of small PR's towards a good solution.

PR #482, opened 2018-10-22
update to .NET standard

MarcBruins:
This PR updates the .NETCoreApp2.0 to .NET standard for the Update and the Inspection project


AnthonySteele:
The Travis build and test run is not working today, but it works on my machine.

PR #487, opened 2018-10-27
Abstractions

shep1987:
I had a few spare hours so working off @MarcBruins work, this is just the Abstractions and GitHubClient move.

All tests are passing. Are we happy with this or is it still a bit too big?

IBranch, IPullRequest, IAccount etc still all use the octokit types, is this OK to come later in a different PR?


NuKeeper.Github/Mappings/Account.cs
```diff
@@ -0,0 +1,20 @@
...
+    public class GithubAccount : User, IAccount
+    {
+
+        public GithubAccount(string avatarUrl, string bio, string blog, int collaborators, string company, DateTimeOffset createdAt, DateTimeOffset updatedAt, int diskUsage, string email, int followers, int following, bool? hireable, string htmlUrl, int totalPrivateRepos, int id, string location, string login, string name, string nodeId, int ownedPrivateRepos, Plan plan, int privateGists, int publicGists, int publicRepos, string url, RepositoryPermissions permissions, bool siteAdmin, string ldapDistinguishedName, DateTimeOffset? suspendedAt)
+            :base(avatarUrl,bio,blog,collaborators,company,createdAt,updatedAt,diskUsage,email,followers,following,hireable,htmlUrl,totalPrivateRepos,id,location,login,name,nodeId,ownedPrivateRepos,plan,privateGists,publicGists,publicRepos,url,permissions,siteAdmin,ldapDistinguishedName,suspendedAt)
+        {
+            
+        }
+
+        public new string Login => base.Login;
```
slang25: You can omit these

NuKeeper.Github/Mappings/Account.cs
```diff
@@ -0,0 +1,20 @@
+using NuKeeper.Abstract;
+using Octokit;
+using System;
+#pragma warning disable CA1054 // Uri parameters should not be strings
+namespace NuKeeper.Github.Mappings
+{
+    public class GithubAccount : User, IAccount
```
slang25: Something seems off creating interfaces for DTOs, are they really needed? It seems like base classes would work just fine here.

NuKeeper.Abstract/Configuration/AuthSettings.cs
```diff
@@ -1,10 +1,10 @@
 using System;
 
-namespace NuKeeper.Configuration
+namespace NuKeeper.Abstract.Configuration
```
slang25: In your comment you referred to this as `Abstractions`, I think `NuKeeper.Abstractions` a more suitable name than `NuKeeper.Abstract` and I've seen it in other libraries, not sure what everyone else feels about this.

NuKeeper.Github/Mappings/AutoMapperConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public static class AutoMapperConfiguration
+    {
+        public static IMapper GithubMappingConfiguration { get; }
+
+        [SuppressMessage("ReSharper", "CA1810")]
+        static AutoMapperConfiguration()
+        {
+            var mapper = new Lazy<MapperConfiguration>(()=>
+                new MapperConfiguration(cfg =>
```
slang25: For this to be actually lazy then I think you need to put the `CreateMapper` call inside of the `Lazy` and the `.Value` part should go in the property getter.

This looks like you are eagerly creating it the first time and setting the static property that value.

AnthonySteele: Why do we have automapper introduced in the first place?

NuKeeper.Github/Mappings/Account.cs
AnthonySteele: I strongly second this comment.  Interfaces are used so that services can be a) injected by typed by the container and b) have behaviour mocked out for testing.  

But objects that are purely data not code are DTOs. DTOs have no behaviour to mock and can be created in tests when needed. They should not have interfaces.

shep1987:
Thanks both, will work on the above today.

AnthonySteele:
Maybe deal with the project organisation first? Then types later.

NuKeeper.Abstract/Configuration/AuthSettings.cs
shep1987: I'm not tied to either... If you would like me to change it i can

NuKeeper.Github/Mappings/Account.cs
shep1987: I have left these as interfaces at the moment. Happy to swap them to base classes though. let me know your thoughts

NuKeeper.Tests/Engine/RepositoryUpdaterTests.cs
```diff
@@ -102,9 +99,11 @@ public async Task WhenThereAreUpdates_CountIsAsExpected(int numberOfUpdates, boo
             Assert.That(count, Is.EqualTo(expectedUpdates));
 
             await gitHub.Received(expectedPrs)
```
shep1987: Self review: Check this

shep1987:
Gentlemen, this is at a point where I would be looking to you both to let me know what else you would like doing.

Thanks for the prompt reviews so far!

NuKeeper.Update/Selection/IFilterSettings.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Text.RegularExpressions;
+
+namespace NuKeeper.Update.Selection
+{
+    public interface IFilterSettings
```
AnthonySteele: Another "interface to DTO" for no reason, and there are others.  How did it come about? I would revert and start again with a simpler change that doesn't include this antipattern. I don't want that on master.

MarcBruins:
Maybe we should close this as well along with mine #480 and keep this as a reference. I've listed an issue #489 which has smaller steps towards the solution of this problem. We should be able to make small PR's faster and make some progres.

shep1987:
@MarcBruins Sounds good to me, have pushed some work i did this morning around using DTO's instead of interfaces

PR #491, opened 2018-10-30
Change IGitHub to use abstractions DTO's

shep1987:
Update IGitHub to use Nukeeper DTOs instead of Octokit

Some changes were made these are as follows:
**In IGithub**
```
Task<PullRequest> OpenPullRequest(ForkData target, NewPullRequest request, IEnumerable<string> labels);
Task OpenPullRequest(ForkData target, PullRequestRequest request, IEnumerable<string> labels);
```
The return type of this method was not used so to keep the number of new classes down i removed it

```
Task<Branch> GetRepositoryBranch(string userName, string repositoryName, string branchName);
Task<bool> RepositoryBranchExists(string userName, string repositoryName, string branchName);
```
The only time we used branch was to check if it was null again to keep new classes down simplified this

**DTOs**
Created new base DTOs in abstractions that accept simple props in the constructor, this is to keep them immutable, we know everything about them when we retrieve them 

Created new GitHub<DTO> classes that have the Octokit type in the constructor, this provides a convenient way to to wrap existing objects to return the new DTOs

At the moment I have tried to keep this PR simple by wrapping all existing objects, in a future PR (or this one if you like)  Updating `RepositoryBuilder` to return the new DTO's would be beneficial

### DIFF
--- a/NuKeeper.Abstractions/DTOs/Repository.cs
+++ b/NuKeeper.Abstractions/DTOs/Repository.cs
@@ -1,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace NuKeeper.Abstractions.DTOs
+{
+    public class Repository
+    {
+        public Repository(string name, bool archived, UserPermissions userPermissions, Uri htmlUrl, Uri cloneUrl, User owner, bool fork, Repository parent)
+        {
+            Name = name;
+            Archived = archived;
+            UserPermissions = userPermissions;
+            HtmlUrl = htmlUrl;
+            CloneUrl = cloneUrl;
+            Owner = owner;
+            Fork = fork;
+            Parent = parent;
+        }
+
+        public string Name { get; }
+        public bool Archived { get; }
+        public UserPermissions UserPermissions { get; }
+        public Uri HtmlUrl { get; }
+        public User Owner { get; }
+        public bool Fork { get; }
+        public Repository Parent { get; }
+        public Uri CloneUrl { get; set; }
+    }
+}
--- a/NuKeeper.Abstractions/DTOs/UserPermissions.cs
+++ b/NuKeeper.Abstractions/DTOs/UserPermissions.cs
@@ -1,0 +1,27 @@
+namespace NuKeeper.Abstractions.DTOs
+{
+    public class UserPermissions
+    {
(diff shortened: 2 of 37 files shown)


AnthonySteele:
I like it! 
I Agree that `RepositoryBuilder` is a helper that can now be drastically simplified, and maybe the " GitHub classes" as well. I would say not in teis PR.

NuKeeper.Abstractions/DTOs/Repository.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace NuKeeper.Abstractions.DTOs
+{
+    public class Repository
+    {
+        public Repository(string name, bool archived, UserPermissions userPermissions, Uri htmlUrl, Uri cloneUrl, User owner, bool fork, Repository parent)
```
MarcBruins: This constructor has a lot of mandatory parameters. Are all of them necessary?

NuKeeper.Abstractions/DTOs/UserPermissions.cs
```diff
@@ -0,0 +1,27 @@
...
+    public class UserPermissions
+    {
+        public UserPermissions(bool admin, bool push, bool pull)
+        {
+            Admin = admin;
+            Push = push;
+            Pull = pull;
+        }
+
+        /// <summary>
```
MarcBruins: These comments are nice and descriptive on our DTO's. We can use them on the other properties for our DTO's as well.

NuKeeper.Tests/Engine/ForkFinderTests.cs
```diff
@@ -30,7 +29,7 @@ public async Task ThrowsWhenNoPushableForkCanBeFound()
         public async Task FallbackForkIsUsedWhenItIsFound()
         {
             var fallbackFork = DefaultFork();
-            var fallbackRepoData = RepositoryBuilder.MakeRepository();
+            var fallbackRepoData = new GitHubRepository(RepositoryBuilder.MakeRepository());
```
MarcBruins: We should split up the tests in the future as well. That will be to much for now probably

AnthonySteele:
Agreed, a better folder name would answer the question "what are these DTOs for", and it might be "Repository".

PR #493, opened 2018-10-30
Move 2 pieces of helper code into Abstractions

AnthonySteele:
Move 2 bits of helper code: `LinqAsync` and `InList` into the abstractions project.

You might say that `Abstractions`  is in danger of becoming a "GeneralFunctions" / "helpers" project and I agree that this is a smell worth keeping an eye on. However I think that it is an improvement - i.e. `LinqAsync` was where it was, not because it was really an "update" function, but because that allowed it to be called from an "update" function, although it is generic infrastructure.  Better to have it with the other "loosely coupled" code pieces.

`InList` was only used in test, but it's small and I think generally useful.

### DIFF
--- a/NuKeeper.Abstractions/LinqAsync.cs
+++ b/NuKeeper.Abstractions/LinqAsync.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 
-namespace NuKeeper.Update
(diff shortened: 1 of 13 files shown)


MarcBruins:
LGTM. I agree that we should keep an eye out before this is getting to big. Maybe will find out a better approach the more we get into abstracting it.

PR #497, opened 2018-10-31
Move octokit to NuKeeper.GitHub

shep1987:
This PR removes the Octokit dependancy from the "NuKeeper" Project and moves it into Nukeeper.Github.

First commit is the bulk of the change, second is renaming everything. There are still bits of GitHub Functionality to move into the new project but i feel this is a good first step.


AnthonySteele:
A good start, I think.
Will still have to look at places where `NuKeeper.GitHub` depends on `NuKeeper.Inspection`

NuKeeper.Tests/Engine/GitHubEngineTests.cs
```diff
@@ -1,13 +1,14 @@
...
+using NuKeeper.Abstractions.DTOs;
 using NuKeeper.Abstractions.Logging;
 using NuKeeper.Engine;
-using NuKeeper.GitHub;
 using NuKeeper.Inspection.Files;
 using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using NuKeeper.Abstractions.CollaborationPlatform;
```
skolima: Is using sorting disabled?

shep1987: looks like when using resharper to refactor it doesn't re-order afterwards... will go through and fix that now

AnthonySteele:
Is this ready to merge now?

PR #500, opened 2018-11-01
Fix for filter not performing correct check

shep1987:
I believe i broke this when editing RepositoryBranchExists.

Added test to ensure that doesn't happen again

### DIFF
--- a/NuKeeper.Tests/Engine/ExistingBranchFilterTests.cs
+++ b/NuKeeper.Tests/Engine/ExistingBranchFilterTests.cs
@@ -24,11 +24,23 @@ namespace NuKeeper.Tests.Engine
         {
             var fork = new ForkData(new Uri("http://uri.com"), "owner", "name");
             var collaborationPlatform = Substitute.For<ICollaborationPlatform>();
-            collaborationPlatform.RepositoryBranchExists(fork.Name, fork.Owner, Arg.Any<string>()).Returns(false);
+            collaborationPlatform.RepositoryBranchExists(fork.Owner, fork.Name, Arg.Any<string>()).Returns(false);
 
             var filter = new ExistingBranchFilter(collaborationPlatform, Substitute.For<INuKeeperLogger>());
             var result = await filter.CanMakeBranchFor(MakeUpdateSet(), fork);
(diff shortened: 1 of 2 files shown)


NuKeeper.Tests/Engine/ExistingBranchFilterTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        public async Task IfBranchDoesNotExistAllowCreation()
+        {
+            var fork = new ForkData(new Uri("http://uri.com"), "owner", "name");
+            var collaborationPlatform = Substitute.For<ICollaborationPlatform>();
+            collaborationPlatform.RepositoryBranchExists(fork.Name, fork.Owner, Arg.Any<string>()).Returns(false);
+
+            var filter = new ExistingBranchFilter(collaborationPlatform, Substitute.For<INuKeeperLogger>());
+            var result = await filter.CanMakeBranchFor(MakeUpdateSet(), fork);
+            Assert.AreEqual(result, true);
+        }
```
AnthonySteele: Can you also do the opposite case - `IfBranchDoesExist DoNotAllowCreation `

NuKeeper/Engine/Packages/ExistingBranchFilter.cs
```diff
@@ -24,7 +24,7 @@ public ExistingBranchFilter(ICollaborationPlatform collaborationPlatform, INuKee
             try
             {
                 var branchName = BranchNamer.MakeSinglePackageName(packageUpdateSet);
-                return await _collaborationPlatform.RepositoryBranchExists(pushFork.Owner, pushFork.Name, branchName);
+                return await _collaborationPlatform.RepositoryBranchExists(pushFork.Owner, pushFork.Name, branchName) == false;
```
AnthonySteele: I'm not a fan of `== false` 
I would even put it on 2 lines if it makes is easier to deal with the await and the conditional, e.g.
```
var branchExists = await _collaborationPlatform.RepositoryBranchExists(pushFork.Owner, pushFork.Name, branchName);
return ! branchExists;
```

PR #502, opened 2018-11-03
Add Azure Devops Project

shep1987:
Looking for feedback on the first steps of adding Azure Devops Support

Added `ISettingsReader` needs to be implemented by each platform, this gets the Repository and Auth settings for it
Commands have this injected into the constructor
Added tests around this for the GitHub Implementation (`GitHubSettingsReader`)

`IForkFinder` become something that each platform must implement 
Azure supports only `ForkMode.SingleRepositoryOnly` at the moment

`IRepositoryDiscovery` become something that each platform must implement 
Azure supports only `ServerScope.Repository` at the moment

Added most of the implementation to `AzureDevOpsPlatform` using a rest client but this will need further work to make it more efficient and improve the structure but it does work.

Removed/Renamed some GitHub Specific language.


NuKeeper.Abstractions/Configuration/AuthSettings.cs
```diff
@@ -12,5 +12,6 @@ public AuthSettings(Uri apiBase, string token)
 
         public Uri ApiBase { get; }
         public string Token { get; }
+        public string AzureDevOpsOrganisation { get; set; }
```
skolima: abstraction leaking?

shep1987: Yup, nice catch fix pushed

AnthonySteele:
Seems OK to me. Is this good to merge now?

shep1987:
I am happy with it

PR #506, opened 2018-11-08
Collaboration factory

shep1987:
Add the start of the collaboration factory,

Have spit this into sections as it was getting too large to submit in one go.

After swapping registrations in `ContainerRegistration` allows you to run:
`repo https://dev.azure.com/<org>/<project>/_git/<repo> <token> --fork=SingleRepositoryOnly`

Only works for new style Azure uri's, support could be added for legacy scheme by implementing a new setting reader in a future PR

**Changes**

- Move auth settings into the CollaborationFactory
- Use an IEnumerable<ISettingReaders>> to find which platform a Uri is for and set the default api accordingly.
- Added a `Platform` enum to be able to easily tell which platform is being used

### DIFF
--- a/NuKeeper.Abstractions/CollaborationPlatform/CollaborationPlatformSettings.cs
+++ b/NuKeeper.Abstractions/CollaborationPlatform/CollaborationPlatformSettings.cs
@@ -6,6 +6,5 @@ namespace NuKeeper.Abstractions.CollaborationPlatform
     {
         public Uri BaseApiUrl { get; set; }
         public string Token { get; set; }
-
     }
 }
--- a/NuKeeper.Abstractions/CollaborationPlatform/CollaborationPlatformSettings.cs
+++ b/NuKeeper.Abstractions/CollaborationPlatform/CollaborationPlatformSettings.cs
@@ -1,0 +1,11 @@
+using System;
+
+namespace NuKeeper.Abstractions.CollaborationPlatform
+{
+    public class CollaborationPlatformSettings
+    {
+        public Uri BaseApiUrl { get; set; }
+        public string Token { get; set; }
+
+    }
+}
--- a/NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
+++ b/NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
@@ -11,7 +11,8 @@ namespace NuKeeper.AzureDevOps
     // https://dev.azure.com/{org}/{project}/_git/{repo}/
     public class AzureDevOpsSettingsReader : ISettingsReader
     {
-        public Platform Platform => Platform.AzureDevOps;
+        private const Platform AzureDevOps = Platform.AzureDevOps;
+        public Platform Platform => AzureDevOps;
 
         public bool CanRead(Uri repositoryUri)
         {
--- a/NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
+++ b/NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
@@ -7,8 +7,6 @@ using NuKeeper.Abstractions.Formats;
 
 namespace NuKeeper.AzureDevOps
 {
-    // URL pattern is
-    // https://dev.azure.com/{org}/{project}/_git/{repo}/
     public class AzureDevOpsSettingsReader : ISettingsReader
     {
         private const Platform AzureDevOps = Platform.AzureDevOps;
@@ -26,7 +24,6 @@ namespace NuKeeper.AzureDevOps
 
         public AuthSettings AuthSettings(Uri apiUri, string accessToken)
         {
-
             if (apiUri == null)
             {
                 return null;
@@ -53,6 +50,8 @@ namespace NuKeeper.AzureDevOps
                 return null;
             }
 
+            // URL pattern is
+            // https://dev.azure.com/{org}/{project}/_git/{repo}/
             var path = repositoryUri.AbsolutePath;
             var pathParts = path.Split('/')
                 .Where(s => !string.IsNullOrWhiteSpace(s))
--- a/NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
+++ b/NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
@@ -1,23 +1,48 @@
 using System;
 using System.Linq;
+using NuKeeper.Abstractions;
 using NuKeeper.Abstractions.CollaborationPlatform;
 using NuKeeper.Abstractions.Configuration;
+using NuKeeper.Abstractions.Formats;
 
 namespace NuKeeper.AzureDevOps
 {
+    // URL pattern is
+    // https://dev.azure.com/{org}/{project}/_git/{repo}/
     public class AzureDevOpsSettingsReader : ISettingsReader
     {
         public Platform Platform => Platform.AzureDevOps;
 
         public bool CanRead(Uri repositoryUri)
         {
-            // general pattern is either of
-            // https://dev.azure.com/{org}/{project}/_git/{repo}/
             if (repositoryUri == null || repositoryUri.Host != "dev.azure.com")
             {
                 return false;
             }
+
             return true;
+        }
+
+        public AuthSettings AuthSettings(Uri apiUri, string accessToken)
+        {
+
+            if (apiUri == null)
+            {
+                return null;
+            }
+
+            var testUri = UriFormats.EnsureTrailingSlash(apiUri);
+
+            if (!testUri.IsWellFormedOriginalString()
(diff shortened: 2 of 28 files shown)


NuKeeper/ContainerRegistration.cs
```diff
@@ -38,15 +38,16 @@ private static void Register(Container container)
             container.Register<IUpdateSelection, UpdateSelection>();
             container.Register<IFileSettingsCache, FileSettingsCache>();
 
+            container.RegisterSingleton<ICollaborationFactory,CollaborationFactory>();
+            container.Collection.Register<ISettingsReader>(typeof(GitHubSettingsReader), typeof(AzureDevOpsSettingsReader));
```
timmkrause: Would it make sense to generically pull `ISettingsReader` implementations via reflection or do you consider this as overengineering? Potentially this would make future integrations of additional platforms more easy.

shep1987: Yup I like the Idea, can we add it to the list of bonus work rather than complicating this PR?

NuKeeper.Abstractions/CollaborationPlatform/CollaborationPlatformSettings.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace NuKeeper.Abstractions.CollaborationPlatform
+{
+    public class CollaborationPlatformSettings
+    {
+        public Uri BaseApiUrl { get; set; }
+        public string Token { get; set; }
+
```
MarcBruins: Empty line

NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
```diff
@@ -7,13 +7,42 @@
 
 namespace NuKeeper.AzureDevOps
 {
+    // URL pattern is
+    // https://dev.azure.com/{org}/{project}/_git/{repo}/
     public class AzureDevOpsSettingsReader : ISettingsReader
     {
-        private readonly FileSettings _fileSettings;
+        public Platform Platform => Platform.AzureDevOps;
```
MarcBruins: This will return a new instance of the enum every time you call the property Platform. This should probably be a constant

NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
```diff
@@ -7,13 +7,42 @@
...
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        public AuthSettings AuthSettings(Uri apiUri, string accessToken)
+        {
+
```
MarcBruins: extra enter

NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
```diff
@@ -23,54 +52,27 @@ public RepositorySettings RepositorySettings(Uri repositoryUri)
...
-            // https://{org}.visualstudio.com/{project}/_git/{repo}/
-            // https://dev.azure.com/{org}/{project}/_git/{repo}/
-            // from this we extract owner and repo name
             var path = repositoryUri.AbsolutePath;
             var pathParts = path.Split('/')
                 .Where(s => !string.IsNullOrWhiteSpace(s))
                 .ToList();
 
-            if (pathParts.Count < 3)
+            if (pathParts.Count != 4)
```
MarcBruins: What kind of magic number is this? Maybe put in a property and have a descriptive name?

NuKeeper.GitHub/GitHubSettingsReader.cs
```diff
@@ -9,34 +9,32 @@ namespace NuKeeper.GitHub
...
-            const string defaultGithubApi = "https://api.github.com/";
-            var api = Concat.FirstValue(apiEndpoint, _fileSettings.Api, defaultGithubApi);
+            if (apiUri == null)
+            {
+                return null;
+            }
 
-            var baseUri = new Uri(api, UriKind.Absolute);
-            baseUri = UriFormats.EnsureTrailingSlash(baseUri);
+            var testUri = UriFormats.EnsureTrailingSlash(apiUri);
```
MarcBruins: Why is this called testUri?

NuKeeper.GitHub/GitHubSettingsReader.cs
```diff
@@ -9,34 +9,32 @@ namespace NuKeeper.GitHub
 {
     public class GitHubSettingsReader : ISettingsReader
     {
-        private readonly FileSettings _fileSettings;
+        public Platform Platform => Platform.GitHub;
```
MarcBruins: Same as above

NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
shep1987: Have move the comment on what this bit of code is trying to do closer, is this is descriptive enough now?

MarcBruins: Ah it makes sense when viewing the arrayparts below

NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
skolima: Tech note: that's not true. A) Enums are (underneath) 32 bit value types,  passed by value and have same 'cost' as passing an e.g. memory address around B) Public constants are dangerous, because they get embedded into compiled user code C) single-op methods are inlined by JIT compiler at runtime anyway, so no need to overo-optimise them.

NuKeeper.GitHub/GitHubSettingsReader.cs
skolima: Exactly, same as above :P

NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
AnthonySteele: We're generally not worried about this kind of optimisation in NuKeeper. Overhead is minimal and the run-time is dominated by the cost of file system access, remote api requests and running external processes.

NuKeeper/Commands/RepositoryCommand.cs
```diff
@@ -1,45 +1,62 @@
...
+                if (reader.CanRead(repoUri))
+                {
+                    didRead = true;
+                    settings.SourceControlServerSettings.Repository = reader.RepositorySettings(repoUri);
+                }
+            }
+
+            if (!didRead)
+            {
+                return ValidationResult.Failure($"Unable to work out which platform to use");
```
AnthonySteele: Same here, the message sound mention the `GitHubRepositoryUri` that could not be matched

NuKeeper/Commands/GitHubNuKeeperCommand.cs
```diff
@@ -32,18 +34,21 @@ internal abstract class GitHubNuKeeperCommand : CommandBase
...
-            Description = "Label to apply to GitHub pull requests. Defaults to 'nukeeper'. Multiple labels can be provided by specifying this option multiple times.")]
+            Description =
+                "Label to apply to GitHub pull requests. Defaults to 'nukeeper'. Multiple labels can be provided by specifying this option multiple times.")]
         public List<string> Label { get; set; }
 
         [Option(CommandOptionType.SingleValue, ShortName = "g", LongName = "api",
-            Description = "GitHub Api Base Url. If you are using an internal GitHub server and not the public one, you must set it to the api url for your GitHub server.")]
+            Description =
+                "GitHub Api Base Url. If you are using an internal GitHub server and not the public one, you must set it to the api url for your GitHub server.")]
         public string GithubApiEndpoint { get; set; }
```
AnthonySteele: We can rename the property from `GithubApiEndpoint` to something else, maybe just `ApiEndpoint` as the commandline option won't be affected.

NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
```diff
@@ -9,11 +9,37 @@ namespace NuKeeper.AzureDevOps
...
+
+            var testUri = UriFormats.EnsureTrailingSlash(apiUri);
+
+            if (!testUri.IsWellFormedOriginalString()
+                || (testUri.Scheme != "http" && testUri.Scheme != "https"))
+            {
+                return null;
+            }
+
+            var envToken = Environment.GetEnvironmentVariable("NuKeeper_azure_devops_token");
```
AnthonySteele: This env var should go in the docs after merge. 
Or is it better to rename the env var to something more general, e.g. `nukeeper_auth_token`

shep1987: After discussion on Gitter will leave as is for now

PR #511, opened 2018-11-11
Move ICollaborationPlatform into factory

shep1987:
Move the final class into the factory

This means you can now run the repo command against GitHub and Azure DevOps
`repo https://dev.azure.com/<org>/<project>/_git/<repo> <token> --fork=SingleRepositoryOnly`


AnthonySteele:
I am catching up with the code idioms here. So now `CollaborationFactory ` is the key to all the platform-dependant things. (no objection, just clarifying)

shep1987:
> I am catching up with the code idioms here. So now `CollaborationFactory ` is the key to all the platform-dependant things. (no objection, just clarifying)

Yup that is correct, using the Initialise method it works out the platform and returns you the correct implementation for each of the interfaces based on the selected platform

PR #522, opened 2018-11-14
Add the Bitbucket platform

MarcBruins:
This PR adds the bitbucket platform. This means that with the REPO command you can now also target bitbucket. In order to do so you need to create a App Password in the portal and use that. The command format is:

`repo https://{username}@bitbucket.org/{username}/{repo} {appPassword}`

I know it's a large PR, but the largest part of it is the 37 POCOS located in the Models directory that are imported.


NuKeeper.BitBucket/BitbucketRepositoryDiscovery.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using NuKeeper.Abstractions.CollaborationPlatform;
+using NuKeeper.Abstractions.Configuration;
+using NuKeeper.Abstractions.Logging;
+
+namespace NuKeeper.BitBucket
+{
+    public class BitbucketRepositoryDiscovery : IRepositoryDiscovery
```
MarcBruins: This file is exactly the same as the AzureDevOps one

NuKeeper.BitBucket/BitbucketRestClient.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        private readonly HttpClient _client;
+        private readonly INuKeeperLogger _logger;
+
+        public BitbucketRestClient(HttpClient client, INuKeeperLogger logger, string username, string appPassword)
+        {
+            _client = client;
+            _logger = logger;
+
+            var byteArray = Encoding.ASCII.GetBytes($"{username}:{appPassword}");
```
MarcBruins: Here we use the Username to create a basic auth login

NuKeeper.BitBucket/BitbucketRestClient.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Net.Http.Headers;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using NuKeeper.Abstractions.Logging;
+using NuKeeper.BitBucket.Models;
+
+namespace NuKeeper.BitBucket
+{
+    public class BitbucketRestClient
```
MarcBruins: Improvement for the clients would be to handle errors if they occur, so log that a request failed

NuKeeper.BitBucket/Models/Activity.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NuKeeper.BitBucket.Models
+{
+    public class Activity
```
MarcBruins: Start of the models

NuKeeper.BitBucket/BitbucketSettingsReader.cs
```diff
@@ -0,0 +1,67 @@
...
+using NuKeeper.Abstractions.CollaborationPlatform;
+using NuKeeper.Abstractions.Configuration;
+
+namespace NuKeeper.BitBucket
+{
+    public class BitbucketSettingsReader : ISettingsReader
+    {
+        public Platform Platform => Platform.Bitbucket;
+
+        private static string Username { get; set; }
```
MarcBruins: So we need the username in our system to compose the basic auth header. We store it in a static string when the `RepositorySettings(Uri repositoryUri)`  is called and then get it out on the `UpdateCollaborationPlatformSettings(CollaborationPlatformSettings settings)`. Those two methods are from two different intances so that's why it's static.

My proposal would be to make all the ISettingsReader singleton when we injected them in the dependency container.

NuKeeper.BitBucket/BitbucketSettingsReader.cs
```diff
@@ -0,0 +1,67 @@
...
+        
+        public bool CanRead(Uri repositoryUri)
+        {
+            if (repositoryUri == null || !repositoryUri.ToString().ToUpperInvariant().Contains("bitbucket.org".ToUpperInvariant()))
+            {
```
AnthonySteele: This can now be `repositoryUri.ToString().ContainsOrdinal("bitbucket.org")`

There's a new helper method on master

NuKeeper.BitBucket/BitbucketPlatform.cs
```diff
@@ -0,0 +1,112 @@
...
+            var repos = await _client.GetGitRepositories(projectName);
+            return repos.Select(repo =>
+                new NuKeeper.Abstractions.DTOs.Repository(repo.name, false, new UserPermissions(true, true, true), new Uri(repo.links.clone.First().href), new Uri(repo.links.clone.First().href), null, false, null))
+                .ToList();
+        }
+
+        public async Task<NuKeeper.Abstractions.DTOs.Repository> GetUserRepository(string projectName, string repositoryName)
+        {
+            var repos = await GetRepositoriesForOrganisation(projectName);
+            return repos.Single(x => x.Name.ToUpperInvariant() == repositoryName.ToUpperInvariant());
```
AnthonySteele: Better to use `string.Equals(a,b, StringComparison.OrdinalIgnoreCase)` for these.

PR #541, opened 2018-11-18
Refactored the CollaborationFactory

AnthonySteele:
Only one `switch` statement on the platform.
This was a difficult class, with properties that trigger actions that trigger the properties again.
I have tried to flatten that,  so that there is a clear "initialise" step and it's populated afterwards.


shep1987:
LGTM

PR #546, opened 2018-11-20
Simplify the org class

AnthonySteele:
Now that we don't use the octokit types throughout, the coalescence, `org.Name ?? org.Login` can move up from `GitHubRepositoryDiscovery` to `OctokitClient`

### DIFF
--- a/NuKeeper.Git/LibGit2SharpDiscoveryDriver.cs
+++ b/NuKeeper.Git/LibGit2SharpDiscoveryDriver.cs
@@ -70,8 +70,8 @@ namespace NuKeeper.Git
         public GitRemote GetRemoteForPlatform(Uri repositoryUri, string platformHost)
         {
             var remotes = GetRemotes(repositoryUri);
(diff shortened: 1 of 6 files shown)


NuKeeper.Git/LibGit2SharpDiscoveryDriver.cs
```diff
@@ -70,8 +70,8 @@ public string GetCurrentHead(Uri repositoryUri)
         public GitRemote GetRemoteForPlatform(Uri repositoryUri, string platformHost)
         {
             var remotes = GetRemotes(repositoryUri);
-            var origin = remotes.FirstOrDefault(rm => rm.Url.Host.ContainsOrdinal(platformHost) == true);
-            return origin;
+            return remotes
+                .FirstOrDefault(rm => rm?.Url?.Host.ContainsOrdinal(platformHost) == true);
```
MarcBruins: 'rm?', it would never be null?

PR #551, opened 2018-11-23
String comparison

MarcBruins:
Add string comparison.

### DIFF
--- a/NuKeeper.Update/Process/UpdateProjectImportsCommand.cs
+++ b/NuKeeper.Update/Process/UpdateProjectImportsCommand.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using System.Xml.Linq;
 using NuGet.Configuration;
 using NuGet.Versioning;
+using NuKeeper.Abstractions.Formats;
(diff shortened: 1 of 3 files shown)


NuKeeper.Update/Process/UpdateProjectImportsCommand.cs
```diff
@@ -54,7 +55,7 @@ private static IEnumerable<string> UpdateConditionsOnProjects(Stream fileContent
 
             var imports = project.Elements(ns + "Import");
             var importsWithToolsPath = imports
-                .Where(i => i.Attributes("Project").Any(a => a.Value.Contains("$(VSToolsPath)")))
```
AnthonySteele: In this case, the text `$(VSToolsPath)` is probably case-sensitive?

skolima: MSBuild files are mostly case-insensitive.

AnthonySteele: That's Ok then.

PR #552, opened 2018-11-23
Replace updated nuspec/csproj files completely when writing

skolima:
Previous write mode would leave characters from the old file behind if the new file was shorter.

### DIFF
--- a/NuKeeper.Update/Process/UpdateProjectImportsCommand.cs
+++ b/NuKeeper.Update/Process/UpdateProjectImportsCommand.cs
@@ -18,21 +18,25 @@ namespace NuKeeper.Update.Process
             var projectsToUpdate = new Stack<string>();
             projectsToUpdate.Push(currentPackage.Path.FullName);
 
-            while(projectsToUpdate.Count > 0)
+            while (projectsToUpdate.Count > 0)
             {
                 var currentProject = projectsToUpdate.Pop();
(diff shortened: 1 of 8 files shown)


NuKeeper.Update/Process/UpdateProjectImportsCommand.cs
```diff
@@ -18,31 +18,34 @@ public class UpdateProjectImportsCommand : IUpdateProjectImportsCommand
             var projectsToUpdate = new Stack<string>();
             projectsToUpdate.Push(currentPackage.Path.FullName);
 
-            while(projectsToUpdate.Count > 0)
+            while (projectsToUpdate.Count > 0)
             {
                 var currentProject = projectsToUpdate.Pop();
+
+                XDocument xml;
                 using (var projectContents = File.Open(currentProject, FileMode.Open, FileAccess.ReadWrite))
```
AnthonySteele: So this shorter "using" and the second "using" in `UpdateConditionsOnProjects` is the important part?

skolima: Yes.

There's 2 file loads now: one to read the project (and close it right away, as the `XDocument` kept in memory is independent. The second load is in truncate mode, and it's optional - only happens when the file needs updating.

PR #553, opened 2018-11-23
String.Contains extension method allowing StringComparison

skolima:
Follows https://docs.microsoft.com/en-us/dotnet/api/system.string.contains?view=netcore-2.1#System_String_Contains_System_String_System_StringComparison_ but works on netstandard2.0

Regarding String.Replace usage in https://github.com/NuKeeperDotNet/NuKeeper/pull/482/files#diff-8a791fc9d6f84fcd97659a3d22de5bb3 : turns out this was only added when the additional code analyzers were introduced, we don't have a bug/use case so far hitting this (and it's quite unlikely).

This method skips the checks on purpose - they're done in `IndexOf` anyhow, and this also keeps exceptions consistent with the netcore2.1 version.


skolima:
Alternative approach to #551

Issue #560, opened 2018-11-24
github remotes not supported


AnthonySteele:
If I run with `repo C:\Code\JustEat.StatsD` this should inspect the local repo and find the remotes.

But I get 

```
Unhandled Exception: System.UriFormatException: Invalid URI: The URI scheme is not valid.
   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
   at System.Uri..ctor(String uriString)
   at NuKeeper.Git.LibGit2SharpDiscoveryDriver.GetRemotes(Uri repositoryUri) in c:\Code\NuKeeper\NuKeeper.Git\LibGit2SharpDiscoveryDriver.cs:line 38
```

This is because there are remotes with e.g. name: "origin" and "url": `git@github.com:AnthonySteele/JustEat.StatsD.git`

And this cannot be cast as `Url` - it is not a http(s) url, although it is a format that github uses a lot.

It's clearly not a simple fix since a run for `repo git@github.com:AnthonySteele/JustEat.StatsD.git` fails similarly.

Do we try to convert these to the corresponding http urls (in this case `https://github.com/AnthonySteele/JustEat.StatsD`) or try to work with them as is?

skolima:
For the remote, the problem is: that's a different git authentication method. Quite likely ssh authentication will work where https doesn't (and vice versa).

AnthonySteele:
Yeah, I think it's a ssh syntax. The question is, if we just feed it back to `LibGit2Sharp`, will it work?

AnthonySteele:
Otherwise the plan would be to manually specify the two separately, e.g. `nukeeper repo --files-are-at  C:\Code\JustEat.StatsD  --remote-is https://github.com/AnthonySteele/JustEat.StatsD`

AnthonySteele:
So if you prefix it with `ssh` , e.g. `ssh:git@github.com:AnthonySteele/JustEat.StatsD.git` then it is a valid `Uri`.

This doesn't make everything work, but it's a start.

PR #561, opened 2018-11-24
[WIP] Git remotes

AnthonySteele:
Git remotes 

Work for #560


AnthonySteele:
I am going to close this PR. Imight try a different approach some time. 
I was aiming at getting the "local repo" to work with github.
We can inspect the folder and ask `libgit2sharp` for the remotes, and the come back successfully as e.g. `origin  git@github.com:AnthonySteele/NuKeeper.git`
This is a URL, but only if you prefix it with `ssh`, e.g.  `ssh:git@github.com:AnthonySteele/NuKeeper.git`

The issue is that `LibGit2sharp` does not support these well: https://github.com/libgit2/libgit2sharp/issues/1422
So there's no point parsing them.

PR #569, opened 2018-11-28
Additional logging

shep1987:
Also log the inner exception if there is one


AnthonySteele:
When I have some time, can consider this as a cross-cutting thing, i.e. always log the inner exception as well.

PR #572, opened 2018-11-30
Can't search forks

AnthonySteele:
Github search doesn't return results for forks. This is a known limitation.
Unfortunately it just returns empty results which looks like is searched and found nothing

As reported here:

>  I've set it up to "repo" a repository on my Github account ( https://github.com/Jens-H-Eriksen/OctoPack ) ,  but it fails to find any C# projects in method "ContainsDotNetProjects"

### DIFF
--- a/NuKeeper.GitHub/OctokitClient.cs
+++ b/NuKeeper.GitHub/OctokitClient.cs
@@ -150,20 +150,38 @@ namespace NuKeeper.GitHub
         public async Task<SearchCodeResult> Search(SearchCodeRequest search)
         {
             CheckInitialised();
+            var forksCount = 0;
+
             var repos = new RepositoryCollection();
             foreach (var repo in search.Repos)
             {
-                repos.Add(repo.owner, repo.name);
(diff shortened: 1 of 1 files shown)


skolima:
My mistake, I should have put it down as a known limitation. In my opinion it's the correct behaviour (obviously my opinion isn't binding :P ). Reason: even if you don't have push rights to the parent repo, you should still point NuKeeper at that, not at the fork, and simply use a different fork mode to have it create branches in your personal fork and PRs from that.

AnthonySteele:
Hm, that might be right, and I can see how skipping these in a org or global mode is the right thing to do,
Bur there is certainly something to fix here, as the error message is that the fork contains no .NET projects, which is not true.

AnthonySteele:
Should this filter apply in `repo` mode? I'm happy for it to apply when looking at multiple repos, but if you specify a particular repo...

skolima:
You're right, it doesn't make sense to filter in single repo mode.

AnthonySteele:
replaced by #575

PR #574, opened 2018-12-01
CollaborationFactory returns ValidationResult

AnthonySteele:
Use `ValidationResult` not exceptions as return for failures in  `CollaborationFactory.Initialise`


shep1987:
LGTM

PR #581, opened 2018-12-03
Add Yml file

shep1987:
Add the build config for Azure DevOps as a yml file. Meaning build config can be version controlled and reviewed before changes.

### DIFF
--- a/.azure-build.yml
+++ b/.azure-build.yml
@@ -1,0 +1,16 @@
+queue:
+  name: Hosted VS2017
+steps:
+- task: DotNetCoreInstaller@0
+  displayName: 'Use .NET Core sdk 2.1.500'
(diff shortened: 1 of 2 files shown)


MarcBruins:
Why do we do a publish here?

shep1987:
image


> Why do we do a publish here?

image
This is the same as we do in the current release. Instead of having 2 configs one for PR and one for Master this just allows us to only publish if the branch being built is master (Via a manual trigger)

AnthonySteele:
it's really nice to to have the pipeline config as a file under source control.
There is also benefit of having 2 different flows for branch build and for a release from master.
But this can be done with a conditional step?

MarcBruins:
:shipit:

Issue #583, opened 2018-12-03
CLI option to allow updating prerelease packages to only stable packages


cswendrowski:
First off, love this library!

Related to https://github.com/NuKeeperDotNet/NuKeeper/issues/175, my team uses prerelease builds of Nuget packages to continue doing work while the Nuget package PR is being reviewed. Once the package PR has been merged, we would like the option to update our existing branches to that stable release without considering upgrading to a newer Prerelease of that package via a parameter such as `--allowPrereleases false`

This is purely a convenience request since we can always manually update our local branches to the non-beta version in the few cases where a newer prerelease has been pushed out.

AnthonySteele:
Hello, thanks for the feedback!

This is doable. We have discussed adding this option  before, but have not had a real reason to do so before now.
FYI, I was considering an enum with three modes: `AllowPrePrelease.Always`, `AllowPrePrelease.Never`, `AllowPrePrelease.FromPreRelease` where "Allow prerelease from prerelease" is the current behaviour and would stay as default. What you are describing is `Never allow prerelease`.

cswendrowski:
I can take a stab at implementing this. It looks like most of the work is there and I just need to hook up a CLI parameter similar to https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper/Commands/CommandBase.cs#L23, correct?

AnthonySteele:
That's correct, yes

AnthonySteele:
If you have any questions along the way, try chat here: https://gitter.im/NuKeeperDotNet/Lobby

PR #588, opened 2018-12-05
#583 Adds prerelease CLI option

cswendrowski:
Implements Issue #583 

I spent a good amount of time trying to get the Unit Tests to support prereleases cleanly and ended up with this implementation, let me know if you have a better approach I should use

### DIFF
--- a/NuKeeper.Inspection.Tests/NuGetApi/PackageVersionTestData.cs
+++ b/NuKeeper.Inspection.Tests/NuGetApi/PackageVersionTestData.cs
@@ -11,30 +11,56 @@ namespace NuKeeper.Inspection.Tests.NuGetApi
 {
     public static class PackageVersionTestData
     {
+        public const string PrereleaseLabel = "prerelease";
+
         public static List<PackageSearchMedatadata> VersionsFor(VersionChange change)
         {
             switch (change)
             {
                 case VersionChange.Major:
                     return NewMajorVersion()
+                        .Concat(ConvertToPrerelease(NewMajorVersion()))
                         .Concat(MinorVersions())
+                        .Concat(ConvertToPrerelease(MinorVersions()))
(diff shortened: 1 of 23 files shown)


AnthonySteele:
Looks good!  will have time for a full review tomorrow, one comment now...

NuKeeper.Inspection/NuGetApi/ApiPackageLookup.cs
```diff
@@ -17,9 +17,21 @@ public ApiPackageLookup(IPackageVersionsLookup packageVersionsLookup)
...
             PackageIdentity package,
             NuGetSources sources,
-            VersionChange allowedChange)
+            VersionChange allowedChange,
+            UsePrerelease usePrerelease)
         {
-            var allowBetas = package.Version.IsPrerelease;
+            bool allowBetas;
+
+            switch (usePrerelease)
```
AnthonySteele: Extract a method here, and then you can do 
` case UsePrerelease.Always:  return true;` etc

NuKeeper.Inspection.Tests/NuGetApi/PackageVersionTestData.cs
```diff
@@ -11,32 +11,58 @@ namespace NuKeeper.Inspection.Tests.NuGetApi
...
+                        .ToList();
 
                 default:
                     throw new ArgumentOutOfRangeException($"Invalid version change {change}");
             }
         }
 
+        private static List<PackageSearchMedatadata> ConvertToPrerelease(List<PackageSearchMedatadata> packages)
+        {
+            var toReturn = new List<PackageSearchMedatadata>();
```
AnthonySteele: Could be LINQ with `packages.Select(p => PackageVersion(...)).ToList()`

PR #593, opened 2018-12-06
Refactor out common regex match

AnthonySteele:
Refactor out common regex match
Now that the `MatchesIncludeExclude` pattern is used in 3 places

### DIFF
--- a/NuKeeper.Abstractions/RegexMatch.cs
+++ b/NuKeeper.Abstractions/RegexMatch.cs
@@ -11,14 +11,14 @@ namespace NuKeeper.Abstractions
                 && !IsExcluded(target, exclude);
         }
 
-        private static bool IsIncluded(string target, Regex regex)
+        private static bool IsIncluded(string target, Regex include)
         {
-            return regex == null || regex.IsMatch(target);
+            return include == null || include.IsMatch(target);
         }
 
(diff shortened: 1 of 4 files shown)


NuKeeper.Abstractions/RegexMatch.cs
```diff
@@ -0,0 +1,24 @@
...
+        private static bool IsIncluded(string target, Regex regex)
+        {
+            return regex == null || regex.IsMatch(target);
+        }
+
+        private static bool IsExcluded(string target, Regex regex)
+        {
+            return regex != null && regex.IsMatch(target);
+        }
+    }
```
slang25: If you want to go expression crazy:
```csharp
public static bool IncludeExclude(string target, Regex include, Regex exclude)
    => IsIncluded(target, include) && !IsExcluded(target, exclude);
private static bool IsIncluded(string target, Regex regex) => regex?.IsMatch(target) ?? true;
private static bool IsExcluded(string target, Regex regex) => regex?.IsMatch(target) ?? false;
```

AnthonySteele: Hm, I'll think on that. I'm not crazy about complex ternary operator or complex one-liner lambda expressions.
But this might be under the line.

Issue #582, opened 2018-12-03
Path too long


mbernard:
Using the repo command there's no way to specify the checkout path so if you have long path you get this exception
```2018-12-03T21:55:26.2138619Z ##[error]Failed on repo Connect NameConflictException : failed to stat 'C:/Users/VssAdministrator/AppData/Local/Temp/NuKeeper/d88a2128699e46e2821ca3b86f86dc80/src/FrontEnd/Montrium.Connect.FrontEnd.WebApi/Spa/src/views/admin-view/list-management-view/contact-organization-view/organization-list-view/organization-list-right.view.tsx': The filename or extension is too long.
```

shep1987:
Hey @mbernard,

Thanks for trying out the tool! There will always come that time when a path is over 260 chars, so i'm just wondering if we can fix that problem instead of specifying a path.

Can i just check which version of windows you are on? Long path support was added in Windows 10 so if it is before that we may need to look into other options.

mbernard:
Thanks for the fast answer.
I'm actually running on an AzureDevOps hosted agent.
It should be running on Windows 10.

I'm all with you though. If we can fix the issues at the OS or Git level it will be much better.

Any thoughs on how we could do it?

mbernard:
@shep1987 
someone is suggesting this

https://stackoverflow.com/a/40843613/3352999

Do you think it can help? I have unfortunately no way to test it

shep1987:
We have been talking about this on Gitter and think it is that LitGit2 doesn't support long paths
https://github.com/libgit2/libgit2/issues/3053

I may look into using the `Agent.TempDirectory` env var (Set by default on the build server) which should be a shorter path. Will have a chat with people tomorrow about this being a viable work around.

For my ref: %tmp% and %temp% should already get overridden by the agent we use Path.GetTempPath() which doesn't seem to respect this..
https://github.com/Microsoft/azure-pipelines-agent/pull/824/files

AnthonySteele:
If there's a way to fix this, we should document it. 

By the way, if you run NuKeeper with an additional parameter, `--verbosity detailed` or `-v d` you will get a lot more details of that.

So it seems overcomplex to work around with an additional commmandline arg (e.g. `--tempFolder c:\some\path`  but if there's an existing environment var that can be used, we can try that!

mbernard:
@AnthonySteele totally agree with you

[...]

PR #596, opened 2018-12-06
Temporary directory path from environment vars

AnthonySteele:
A workaround for #582 Path too long.
By setting the environment variable `temp`  to a folder with a shorter path, you can gain more chars for the checkout.

In most scenarios, the var `temp` is either not set (linux) or set to the same value as `Path.GetTempPath()` anyway (my windows machine) i.e. `C:\Users\ANTHONY.STEELE\AppData\Local\Temp` so you can make this shorter by pointing it at `c:\temp` or the like.

* We only need one folder factory. It's state doesn't change during the run
*  Get the system temp path at startup
* Check the env var `temp`.

Do not merge yet, let's check if this is the correct environment var to use.

### DIFF
--- a/NuKeeper.Inspection/Files/FolderFactory.cs
+++ b/NuKeeper.Inspection/Files/FolderFactory.cs
@@ -15,7 +15,17 @@ namespace NuKeeper.Inspection.Files
         public FolderFactory(INuKeeperLogger logger)
         {
             _logger = logger;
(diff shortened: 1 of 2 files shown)


AnthonySteele:
Not needed, this env var check replicates [what `Path.GetTempPath` does anyway](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettemppath?view=netstandard-2.0)

Issue #595, opened 2018-12-06
--age 0 filters out packages from the future


cswendrowski:
Here's a fun one from the world of "corporate America has dumb programming problems"

Our internal VM's are known to experience time-drift from each other and are often up to 15 minutes different (although are usually 5 minutes or less). This problem is not unique to self-hosted VM's however - Azure VM's and App Services are known to differ by a few seconds.

Due to this, we have encountered a situation where a freshly deployed package reports as being published in the future when our CI attempts to update our Repos with the new package and is therefore being filtered out.

Relevant output:
```
[Step 4/4] My.Super.Cool.Package to X.Y.Z.P from X.Y.Z.P-1 in 6 places since now.
. . . 
[Step 4/4] Filtered by Include 'My.Super.Cool.Package' from 54 to 1
[Step 4/4] Filtered by minimum package age 'now' from 1 to 0
```

Resolving this might be as simple as setting `_maxPublishedDate` (https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper.Update/Selection/UpdateSelection.cs#L31) to the max `DateTime` if age is 0, which is a tad hacky but would work.

Alternatively a new `-1` option for age could be introduced that would indicate "I want to live on the extreme edge and accept future packages"

This issue can be worked around by delaying CI for a bit before running it to allow the package to slip into "the past"

AnthonySteele:
Haha, from what I've heard, corporates in other countries are not all that different.  Some companies make sure to recycle cloud VMs every day or every week for these clock-drift reasons. I've seen it in AWS.

So zero age was intended to not filter at all - in the usual case on the public nuget feed case anything less than a few hours old is brand new, just arrived.and zero will do that.

It's an interesting case that NuKeeper is just immediately after producing the package. 

how about making `_maxPublishedDate` a `DateTime?` and initialising with:
```
_maxPublishedDate = (settings.MinimumAge == TimeSpan.Zero) ? 
  null : 
  DateTime.UtcNow.Subtract(settings.MinimumAge);
```

an then only applying a filter if it's not null.

cswendrowski:
That seems like a valid approach for handling --age 0 for this edge case

I'm guessing the usual use-case of NuKeeper is running it daily / weekly and checking all Packages or maybe MyCompany.* packages. We debated using that workflow but decided against it due to our high frequency of updates in certain packages and the wide spread of packages we didn't want to auto-update. We currently are using the following workflow:

[...]

PR #597, opened 2018-12-06
Handles packages published in the future relative to NuKeeper

cswendrowski:
#595 

The suggested `?` check wouldn't compile due to `Type of conditional expression cannot be determined because there is no implicit conversion between '<null>' and 'DateTime'` which struck me as odd since I was setting it to a `DateTime?`. I opted for a slightly different approach to the check


AnthonySteele:
Ah, the ternary is strict about both cases having the same declared type - you can do the `?` check with an explicitly typed null: `(DateTime?)null` but the way that you did it is fine.

PR #599, opened 2018-12-07
Handling Directory.Build.props/targets

### LINKED ISSUE #478: Read Directory.Build.targets?
I'm using the approach detailed in this blog post : https://www.strathweb.com/2018/07/solution-wide-nuget-package-version-handling-with-msbuild-15/

I have a Directory.Build.targets file at the root of my repo that contains PackageReferences with a Version attribute set: `<PackageReference Update="Foo" Version="1.0.0" />`

I then have sub-directories that contain solutions with projects etc, inside each of the .csproj files I have PackageReferences without the Version attribute set:  `<PackageReference Include="Foo" />`

This allows the update of a PackageReference in the root file and the change is then inherited in all csproj files that have the reference.

I've read the documentation, but I can't seem to figure out how to call NuKeeper to have it read this file instead of the individual csproj files. Am I doing something wrong?

skolima:
Handles basic scenarios of `Include` (which NuKeeper uses) and `Update` (which was used in the issue created.

Sample PR for Include: https://github.com/skolima/NuKeeper/pull/11/files
Sample PR for Update: https://github.com/skolima/NuKeeper/pull/10/files

Resolves #478

### DIFF
--- a/NuKeeper.Inspection/RepositoryInspection/DirectoryBuildTargetsReader.cs
+++ b/NuKeeper.Inspection/RepositoryInspection/DirectoryBuildTargetsReader.cs
@@ -1,0 +1,81 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+using NuKeeper.Abstractions;
+using NuKeeper.Abstractions.Logging;
+
+namespace NuKeeper.Inspection.RepositoryInspection
+{
+    public class DirectoryBuildTargetsReader : IPackageReferenceFinder
+
+    {
+        private readonly INuKeeperLogger _logger;
+
+        public DirectoryBuildTargetsReader(INuKeeperLogger logger)
+        {
+            _logger = logger;
+        }
+
+        public IReadOnlyCollection<PackageInProject> ReadFile(string baseDirectory, string relativePath)
+        {
+            var packagePath = new PackagePath(baseDirectory, relativePath, PackageReferenceType.DirectoryBuildTargets);
+            try
+            {
+                using (var fileContents = File.OpenRead(packagePath.FullName))
+                {
+                    return Read(fileContents, packagePath);
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new NuKeeperException($"Unable to parse file {packagePath.FullName}", ex);
+            }
+        }
+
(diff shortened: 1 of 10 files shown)


NuKeeper.Update/Process/UpdateDirectoryBuildTargetsCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            using (var xmlOutput = File.Open(currentPackage.Path.FullName, FileMode.Truncate))
+            {
+                UpdateNuspec(xmlOutput, newVersion, currentPackage, xml);
+            }
+
+            return Task.CompletedTask;
+        }
+
+        private void UpdateNuspec(Stream fileContents, NuGetVersion newVersion,
```
slang25: Looks like this was a hangover from using the Nuspec command as a template. Should be something like `UpdateDirectoryBuildTargets` (or something)

NuKeeper.Inspection/RepositoryInspection/DirectoryBuildTargetsReader.cs
```diff
@@ -0,0 +1,81 @@
...
+            var xml = XDocument.Load(fileContents);
+
+            var packagesNode = xml.Element("Project")?.Element("ItemGroup");
+            if (packagesNode == null)
+            {
+                return Array.Empty<PackageInProject>();
+            }
+
+            var packageNodeList = packagesNode.Elements()
+                .Where(x => x.Name == "PackageReference");
```
MarcBruins: Will the target, build file only work with those exact definitions? So only "PackageReference" and not "packageReference" for example?

skolima: It's XML, after all. Yes it's case sensitive. Because of MSBuild quirks (and dotnetcore history), one could also use nodes instead of attributes for `Version`, but no sample I've seen has done that, so I'm only sticking to the current preferred syntax.

skolima: To expand a bit: this file is a normal MSBuild script, in the end, so... it's pretty much free form. We can only rely on the convention to handle the most common cases.

MarcBruins: Great

PR #600, opened 2018-12-07
Relative path support

MarcBruins:
This adds relative path support for when using the repo command.
For example: `nukeeper repo ../ {token}` or `nukeeper repo . {token}`

In the end it will just get the directory that you are calling it from and concats the arguments to make a new absolute path.


AnthonySteele:
Will this affect other cases, e.g. would `nukeeper repo https://github.com/NuKeeperDotNet/NuKeeper <token>`  still work?

MarcBruins:
Yes, that will still work exactly the same as it works now

PR #610, opened 2018-12-13
Use NugetVersion in build

AnthonySteele:
I'm not quite sure that this will work, but in the Azure DevOps build there is a variable called `NugetVersion` which will contain the correct version metadata and we want it to show up here, and not the `VersionPrefix` which is no longer correct.

### DIFF
--- a/.azure-build.yml
+++ b/.azure-build.yml
@@ -3,6 +3,13 @@ queue:
 variables:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
 steps:
+  - bash: echo Set $NugetVersion
+    env:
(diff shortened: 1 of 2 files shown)


.azure-build.yml
```diff
@@ -3,6 +3,13 @@ queue:
 variables:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
 steps:
+  - bash: echo Set $NugetVersion
+    env:
+      NugetVersion: $(NugetVersion)
+  - powershell: Write-Host Set $NugetVersion
```
MarcBruins: We can just use 1, so i would say only bash

PR #611, opened 2018-12-13
Add the VersionPrefix back in

AnthonySteele:
Reinstate `<VersionPrefix>$(NugetVersion)</VersionPrefix>`


AnthonySteele:
For context see https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/

We use this version in `program.cs`

PR #626, opened 2018-12-27
 Target NetCoreApp2.2 not 2.1

AnthonySteele:
Update to the December SDK release
- Target `netcoreapp2.2` not `2.1`
- SDK Version `2.2.101`
- `TreatWarningsAsErrors` and `LangVersion` settings are not needed in `.csproj` files as these are set globally in `Directory.Build.props`

### DIFF
--- a/.azure-build.yml
+++ b/.azure-build.yml
@@ -4,9 +4,9 @@ variables:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
 steps:
 - task: DotNetCoreInstaller@0
-  displayName: 'Use .NET Core sdk 2.1.500'
+  displayName: 'Use .NET Core sdk 2.2.100'
   inputs:
-    version: 2.1.500
+    version: 2.2.100
 
 - bash: echo NugetVersion is $NugetVersion; export NugetVersion
   displayName: 'export NugetVersion into env vars'
--- a/.azure-build.yml
+++ b/.azure-build.yml
@@ -4,9 +4,9 @@ variables:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
 steps:
 - task: DotNetCoreInstaller@0
-  displayName: 'Use .NET Core sdk 2.2.100'
+  displayName: 'Use .NET Core sdk 2.2.101'
   inputs:
-    version: 2.2.100
+    version: 2.2.101
 
 - bash: echo NugetVersion is $NugetVersion; export NugetVersion
   displayName: 'export NugetVersion into env vars'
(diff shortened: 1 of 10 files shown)


AnthonySteele:
`The command "sudo apt-get install -qq dotnet-sdk-2.2=2.2.100*" failed and exited with 100 during .`

This package definitely exists for Ubuntu (I have it on WSL), but doesn't appear to be present on travis.

.azure-build.yml
```diff
@@ -4,9 +4,9 @@ variables:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
 steps:
 - task: DotNetCoreInstaller@0
-  displayName: 'Use .NET Core sdk 2.1.500'
+  displayName: 'Use .NET Core sdk 2.2.100'
   inputs:
-    version: 2.1.500
+    version: 2.2.100
```
MaxMommersteeg: Latest SDK version is 2.2.101 iirc.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: csharp
 mono: none
 dist: trusty
```
MaxMommersteeg: Builds with `trusty` as dist will currently fail. You could use `xenial` temporary: https://travis-ci.community/t/dotnet-core-2-2/1216

AnthonySteele: It looks like Xenial is a later release https://wiki.ubuntu.com/Releases
should probably bump up to Xenial (`16.0`) or Bionic (`18.0`) anyway.
Ubuntu 18 is what I run locally in WSL anyway.

AnthonySteele: Tried `bionic`. It's up to date and t works, so we're good :)

skolima:
The change looks good, my concern would be: what's the scenario now when someone does `dotnet tool install -g nukeeper` on SDK 2.1? Do they get the older version, or does it fail?

AnthonySteele:
Good questions, and that's part of the reason why we did not take an update to NetCore `2.2.x` immediately (unlike updates within `2.1.x`. But it's a question of when not if.

skolima:
For now, the VS installers are *not* including the 2.2 SDK yet. I'd prefer we wait for this to roll out.

AnthonySteele:
Do not merge yet. Will reassess later in January 2019, if VS and the ecosystem are ready for NetCore 2.2 yet.

MarcBruins:
We can also multi target. For example target 2.1 2.2 and 3.0.

AnthonySteele:
It should not be necessary for the commandline app to multi-target NETCore `2.1` and `2.2`.
Will look at moving to 2.2 in the next month, when it is more widespread.

MarcBruins:
But that will require people to have `2.2` installed locally right? With the multi target approach it will work for both `2.1` and `2.2`

Issue #590, opened 2018-12-06
Argument: "Platform"


Jens-H-Eriksen:
As the number of platforms Nukeeper are supporting, are growing, I think It would be a good idea to add an argument, that tells Nukeeper which platform it should use instead of using the logic in the implementations of the CanRead method in ISettingsReader:

`nukeeper <platform> <command> command-arguments options`

AnthonySteele:
I would put it in as an optional argument e.g. `--platform github` for cases where the inference fails.

MarcBruins:
Optional argument sounds a good way forward. I actually like that NuKeeper figures out what platform you are using.

MarcBruins:
Think we should still document this @AnthonySteele?

AnthonySteele:
It should be documented now

ElanHasson:
I'm trying to use the new `platform` argument from the command line and it doesn't appear to work.

On inspection of the command, it doesn't seem to care about it..

https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper/Commands/RepositoryCommand.cs#L49

ElanHasson:
I've found the issue and will be sending a PR.

PR #628, opened 2018-12-27
Started on platform from settings override

AnthonySteele:
Started on platform from settings override
Fix for  #590


AnthonySteele:
Suggestions for a suitable single letter for this option on the command line?

Issue #627, opened 2018-12-27
Repeated out of memory exception


Richiban:
On our build agent we're getting very frequent OutOfMemoryExceptions when running NuKeeper in repo mode. It seems to occur more or less frequently depending on which solution we run it on.

Here's an example: 

```
[12:07:02][Step 5/5] Found NuGet.exe: C:\Users\ContainerAdministrator\.dotnet\tools\.store\nukeeper\0.11.0\nukeeper\0.11.0\tools\netcoreapp2.1\any\NuGet.exe
[12:07:02][Step 5/5] In path C:\BuildAgent\work\98739dabf3b446bd\FunctionalTests, running command: C:\Users\ContainerAdministrator\.dotnet\tools\.store\nukeeper\0.11.0\nukeeper\0.11.0\tools\netcoreapp2.1\any\NuGet.exe update packages.config -Id XXXXXXXX.Sql -Version 1.23.101 -Source https://proget.services.kingsway.asos.com/nuget/ASOSPackages -Source https://proget.services.kingsway.asos.com/nuget/ASOS -Source https://XXXXXXXX.pkgs.visualstudio.com/_packaging/AsosMarketplace/nuget/v2 -Source https://api.nuget.org/v3/index.json
[12:14:41][Step 5/5] Updates failed OutOfMemoryException : Insufficient memory to continue the execution of the program.
[12:14:42][Step 5/5] Error message is logged
[12:14:41][Step 5/5]    at System.Text.StringBuilder.ExpandByABlock(Int32 minBlockCharCount)
[12:14:41][Step 5/5] Attempted 3 updates and did 0
[12:14:41][Step 5/5] Done at 2018-12-27T12:14:41Z
[12:14:41][Step 5/5]    at System.Text.StringBuilder.Append(Char* value, Int32 valueCount)
[12:14:41][Step 5/5]    at System.Text.StringBuilder.Append(Char[] value, Int32 startIndex, Int32 charCount)
[12:14:41][Step 5/5]    at System.IO.StreamReader.ReadToEndAsyncInternal()
[12:14:41][Step 5/5]    at NuKeeper.Update.ProcessRunner.ExternalProcess.Run(String workingDirectory, String command, String arguments, Boolean ensureSuccess) in /home/vsts/work/r1/a/drop/NuKeeper.Update/ProcessRunner/ExternalProcess.cs:line 47
[...]

PR #629, opened 2018-12-27
Nuget update commandline is NonInteractive

AnthonySteele:
Use the `-NonInteractive` flag on update
docs https://docs.microsoft.com/en-gb/nuget/tools/cli-ref-update

For #627


skolima:
The other NuGet command we use is restore, has the same option available https://docs.microsoft.com/en-gb/nuget/tools/cli-ref-restore . It's probably a good idea to set this, in https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper.Update/Process/NuGetFileRestoreCommand.cs

Issue #466, opened 2018-10-10
version ranges


AnthonySteele:
With one of the projects that I target, I get output

```text
Skipping package 'AWSSDK.Core' with version '3.3.*' that could not be parsed.
Skipping package 'AWSSDK.SimpleNotificationService' with version '3.3.*' that could not be parsed.
Skipping package 'AWSSDK.SQS' with version '3.3.*' that could not be parsed.
Skipping package 'AWSSDK.Core' with version '3.3.*' that could not be parsed.
```

Using types in the `NuGet` client libs, we could read these version ranges, and maybe even do something with them?

AnthonySteele:
A start would be this: NuKeeper currently emits that message as it reads the relevant package version line from a file. Instead, it should parse it into the appropriate version-range data structure, and the error can be generated a bit higher up the call stack.

AnthonySteele:
I had a look. we use the type `NuGetVersion` (from the library, in `NuGet.Packaging.Core`) to identify  a specific version. And `PackageIdentity` to identify a package. This consists of `string id` and  `NuGetVersion version`. This type is also in `NuGet.Packaging.Core`. 

So we should use `VersionRange` when reading a version from the file as technically it is one of these, the "specific version" is actually just a special base (and implies a range with only a lower bound). And then collapse to a specific version only when necessary. This would mean  that we discard more complex version ranges much later in the process.

There's no package id + VersionRange equivalent type that I can find, but it's not mpossible to create one.

PR #630, opened 2018-12-28
Package version ranges

AnthonySteele:
I would like the code to work better with version ranges. This is a start.

The package version in a  `.csproj` or `packages.config` file is actually a version range. The single version (e.g. `1.2.3` ) is actually a very common special case. But there are others, e.g. e.g.  `AWSSDK.Core` version `3.3.*`.

We use the library class `NuGet.Versioning.VersionRange` for a range. 

Where the library class `PackageIdentity` has a package Id and a package version, there is no library class to store an Id and Package version range, so I made `PackageVersionRange` .

This is currently converted to a `PackageIdentity` immediately, and drop the items where it can't be converted.  All this gives us is more specific messages about why we can't update that package.

Next steps would be to keep it for longer and convert at the last possible time.

#466

### DIFF
--- a/NuKeeper.Abstractions.Tests/NuGet/VersionRangesTests.cs
+++ b/NuKeeper.Abstractions.Tests/NuGet/VersionRangesTests.cs
@@ -1,0 +1,73 @@
+using NuGet.Versioning;
+using NuKeeper.Abstractions.NuGet;
+using NUnit.Framework;
+
+namespace NuKeeper.Abstractions.Tests.NuGet
+{
+    public class VersionRangesTests
+    {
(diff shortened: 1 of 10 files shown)


AnthonySteele:
This is step one, and gives us nothing but slightly better error messages in the cases where package versions are actually ranges. 

Step 2 would be to use the range in `PackageInProject`,  i.e. make the call to `SingleVersionIdentity()` much later in the whole process.

Step 3 means considering ranges when looking for updates. e.g. what action should be taken when package `Foo` is used at version `2.1.*` and there is an update available to `2.2.1` ?

Issue #568, opened 2018-11-27
Ability to hook in commands


alexanderwilliamson:
Hey, loving the good work. I'm looking to try and get this running with our workflow.

Right now each repository contains a file which denotes the semver. It looks something like this.
```
major: 1
minor: 2
patch: 0
(etc)
```
Each time we create a PR we have the ability to update this file. Most of the time it's always updated, but if you're making a minor change, or working together there's no reason to update the semver. There is a ruby gem we run along the lines of `semver inc minor` which increments the versioning.

## Requirement

When NuKeeper runs, I'd like a way to be able to run a command (like updating that file in the repo) before pushing the files up to a new PR.

## Proposal 1 - The quick-n-dirty workaround - Execute custom script before PR creation

All I can see for now is running `nukeeper repo (etc)` but this command does not let me call out to change this file, or hook in another command. Does anyone have suggestions for this? To move me along I could put a prebuild task in to increment the build then as the `nukeeper repo` command runs it could call out to a .bat or .pr1 or whatever. I understand you are trying to keep this cross platform.

Changes here:
- Ability to hook into and run random file after git pull, and before git PR creation

## Proposal 2 - Update `nukeeper update` and create new command to push

Another approach I could think of would be to have a `nukeeper update` command do the update in the local file system, then I update the semver (through whatever means), then have a new command similar to `nukeeper push`. This would need a few changes. Right now the PR comment cannot be extracted, and is different from the `-o` output file from the `nukeeper update` command. It would be awesome if that could be output to a file by command line.

- `nukeeper update` would ideally create a .md file of the changes
- the ability to step out of the NuKeeper lifecycle and hook something in here (.bat, .ps1, .sh, etc.)
- A new command like `nukeeper push` to do the work the push part of `nukeeper repo` command does, taking in the `.md` file for changes

I hope that makes sense!

AnthonySteele:
Proposal one is very possible, Running commands is something that NuKeeper has done right from the start -  it shells out to `dotnet` and/or `nuget.exe`. Getting a good design is important though. I have 2 suggestions here:

* **Convention** NuKeeper could locate files in the root of the repo that match a certain spec and execute them just before making the PR.  e.g. `NuKeeperBeforePullrequest.(bat|ps1|sh)`. The problem is, what if you have one for linux and one for windows? It's hard to tell which should be run. it's not simple, there is powershell on linux and bash on windows. 

[...]

PR #631, opened 2018-12-30
BeforePullRequestCommand

AnthonySteele:
Hacking around with a "BeforePullRequestCommand"
for #568 

- To generalise this, there could be different places to run commands, that can be hooked into, e.g. "after checkout", "after branch", but not try to cover all the cases, I'll go with YAGNI until someone has a use for that.

- The command is run in the checkout folder, meaning that if the script is in the source repo it will be there with the rest of the repo files. Since the script won't necessarily know what this path is, it is possible to give the command a variable substitution, e.g. replace `{{CheckoutFolder}}` with this path, `{{GitOrigin}}` with that, etc. Unclear if anything like this is needed.

### DIFF
--- a/NuKeeper.Abstractions/Configuration/UserSettings.cs
+++ b/NuKeeper.Abstractions/Configuration/UserSettings.cs
@@ -21,5 +21,6 @@ namespace NuKeeper.Abstractions.Configuration
         public string OutputFileName { get; set; }
 
         public string Directory { get; set; }
+        public string BeforePullRequestCommand { get; set; }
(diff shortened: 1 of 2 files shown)


AnthonySteele:
@3686 does this look like what you need? If so I will flesh it out.

3686:
@AnthonySteele Yes I think it would cover the case I was chatting where we need to make a change to a file in the local directory (plus staging the file if required for the upcoming PR) before PR gets created 👍

Issue #633, opened 2019-01-02
Repository Urls duplication


AnthonySteele:
The class `NuKeeper.Abstractions.CollaborationModels.Repository` has both `CloneUrl` and `HtmlUrl` (both of type URI).

We are only set up to clone using libgit2sharp from http(s) urls, not from ssh urls.

We should try to clarify which one should be used when. And if they overlap, could even remove one?

As noticed in #632

MaxMommersteeg:
This is fixed right?

PR #637, opened 2019-01-06
We only need one url for a repo

AnthonySteele:
We only need one url for a repo. The difference between `HtmlUrl` and `CloneUrl` is not clear, I think it's duplication. #633

Needs testing with github


AnthonySteele:
The issue may be that when matching the git url, it exists in 2 forms that need to be matched with each other
on the browser url bar it is e.g. `https://github.myco.com/SomeTeam/SomeApi`
if you "clone with https" it is  `https://github.myco.com/SomeTeam/SomeApi.git`

We should consider these equivalent, but that should be done in the github handling code, not in the source-control-platform-independent `Repository` class

PR #639, opened 2019-01-10
Maxupdate has 0 option to allow all

MarcBruins:
For a project that i've been working on i use the following command:

`nukeeper update --include ^EventSourcing. --age 0 --maxupdate 50`

This yields about 50 packages update in total. 7 project with about 7 `EventSourcing.*` packages. It would be nice to just say `--maxupdate 0` and allow it to update as many as it wants.

### DIFF
--- a/NuKeeper.Update.Tests/UpdateSelectionTests.cs
+++ b/NuKeeper.Update.Tests/UpdateSelectionTests.cs
@@ -104,7 +104,7 @@ namespace NuKeeper.Update.Tests
             var results = await target.Filter(updateSets, settings, Pass);
 
             Assert.That(results.Count, Is.EqualTo(2));
-            Assert.That(results.First().SelectedId, Is.EqualTo("bar"));
+            Assert.That(results.First().SelectedId, Is.EqualTo("foo"));
         }
 
         [Test]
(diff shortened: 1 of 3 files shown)


AnthonySteele:
Should we use a nullable `int?` for this?

MarcBruins:
So then the default if not `1` anymore but always max?

MarcBruins:
Can you elaborate your comment @AnthonySteele and answer my question?

AnthonySteele:
It should fail safe, i.e. to a small number. 
Can you not  use a larger limit, `500` or `10000` if you want no effective limit?

MarcBruins:
Yes that is what we do now. But that isn't a very nice solution imo. We just want it to update the whole project and we don't know how many updates it has. 

If we could pass in `0` we can always cancel it by just pressing ctrl+c. That is a rather abrupt stop and it doesn't revert changes but we don't care cause we can just revert the changes with git.

In the end it's effectively the same as passing in `500` or `10000` but just a cleaner solution.

AnthonySteele:
We have used `--maxpr 0` to mean "zero updates" so it would be odd to make `0` mean "all the updates" in a different context.

MaxMommersteeg:
Why not allow a set of strings for input? "max" or "all" could be used for all the updates. "none" could be an alias to 0. Uints could easily be parsed.

MarcBruins:
Good suggestion @MaxMommersteeg, i'll look into that.

Issue #624, opened 2018-12-24
stops at the first branch conflict


OneCyrus:
nukeeper (azure devops extension) completely stops after first branch conflict. in this scenario there was just one branch which was already there. nukeeper didn't try the other two dependencies and just failed with ```Attempted 3 updates and did 0``` even though it didn't try 3. imho it should try them all so we can have always 3 PRs open at the same time.

```
2018-12-24T14:39:28.8798353Z Found 4 package updates
2018-12-24T14:39:28.8831409Z Microsoft.Spatial to 7.5.3 from 7.5.0 in 1 place since 5 days ago.
2018-12-24T14:39:28.8831893Z WindowsAzure.Storage to 9.3.3 from 9.3.1 in 1 place since 27 days ago.
2018-12-24T14:39:28.8832349Z MSTest.TestAdapter to 1.4.0 from 1.3.2 in 1 place since 28 days ago.
2018-12-24T14:39:28.8832708Z MSTest.TestFramework to 1.4.0 from 1.3.2 in 1 place since 28 days ago.
2018-12-24T14:39:28.9043280Z Filtered by minimum package age '7 days ago' from 4 to 3
2018-12-24T14:39:28.9890523Z Selection of package updates: 4 candidates, filtered to 3
2018-12-24T14:39:28.9908192Z Selected package update of MSTest.TestAdapter to 1.4.0
2018-12-24T14:39:28.9908687Z Selected package update of MSTest.TestFramework to 1.4.0
2018-12-24T14:39:28.9912177Z Selected package update of WindowsAzure.Storage to 9.3.3
2018-12-24T14:39:29.0091285Z Updating'MSTest.TestAdapter' from 1.3.2 to 1.4.0 in 1 projects
2018-12-24T14:39:29.0554432Z Updates failed NuKeeperException : Git Cannot checkout new branch: a branch named 'origin/nukeeper-update-MSTest.TestAdapter-to-1.4.0' already exists
2018-12-24T14:39:29.0554898Z Attempted 3 updates and did 0
```

MarcBruins:
I agree that it should continue when the branch already exists. Makes sense if you have a scheduled, let say daily job running and you forgot to update one day then the rest of it also fails, untill you merge and/or delete that branch.

AnthonySteele:
Yes, it should try still to create the other updates. This should work on github by checking upfront for a remote branch. But it might be failin on ADO. Which platform are you using?

[...]

PR #650, opened 2019-01-18
No errors when package is not found

MarcBruins:
This PR includes 2 fixes.

1. NuKeeper throws if the branch already exists. This shouldn't happen in my opinion. If you are to lazy to "accept" the PR's then that should be fine. It is also optional in Azure DevOps to delete the branch after a merge. If you keep the branch then NuKeeper will stop working because it detect a stale old branch(shouldn't happen that much because it's bound to a specific upgrade but still).

The `RunFinderForSource` gets called for every nuget source that you have. If it can't authenticate for a specific source that shouldn't be a problem because it can still succeeded, the error create the perception that it fails. 

2. For the project i'm working on right now we don't want any secrets in our code. That means that for some project in our solution we include a `nuget.config` that has credentials that looks like this:

```
  <packageSourceCredentials>
    <VSTS>
      <add key="Username" value="dontcare@dontcare.com" />      
      <add key="ClearTextPassword" value="__System.AccessToken__" />
    </VSTS>
  </packageSourceCredentials>

```
The `System.AccessToken` is injected by our CI. However NuKeeper will use this for it's command. The tool will still work fine because on our machine we have a system wide `nuget.config` which contains a PAT. The downside is that any command generates a hell of a lot `Errors` and our whole screen is almost red.

### DIFF
--- a/NuKeeper.Git/LibGit2SharpDriver.cs
+++ b/NuKeeper.Git/LibGit2SharpDriver.cs
@@ -36,11 +36,11 @@ namespace NuKeeper.Git
             _gitCredentials = gitCredentials;
             _identity = userIdentity;
         }
-        
+
         public void Clone(Uri pullEndpoint)
         {
             _logger.Normal($"Git clone {pullEndpoint} to {WorkingFolder.FullPath}");
-          
+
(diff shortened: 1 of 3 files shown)


AnthonySteele:
The branch part addresses #624 I think.

PR #652, opened 2019-01-18
Include and exclude regex are used early on

MarcBruins:
This PR adds the `include` and `exclude` regex early on in the proces. This is a performance boost, what now happens is that the regex filtered the packages all the way at the end of the proces. With this PR it is done in the beginning, this means that packages are not relevant will not check for updates from potentialy multiple remotes.

Log now reads:
```
Found 44 packages
Filtered by Include '^EventSourcing.' from 44 to 9
Found 9 packages in use, 5 distinct, in 3 projects.
EventSourcing.Extensions, EventSourcing.LeaderElection, EventSourcing.PubSub, EventSourcing.Shared, EventSourcing.Store
```

### DIFF
--- a/NuKeeper.Git/LibGit2SharpDriver.cs
+++ b/NuKeeper.Git/LibGit2SharpDriver.cs
@@ -36,11 +36,11 @@ namespace NuKeeper.Git
             _gitCredentials = gitCredentials;
             _identity = userIdentity;
         }
-        
+
         public void Clone(Uri pullEndpoint)
         {
             _logger.Normal($"Git clone {pullEndpoint} to {WorkingFolder.FullPath}");
-          
+
             Repository.Clone(pullEndpoint.AbsoluteUri, WorkingFolder.FullPath,
                 new CloneOptions
                 {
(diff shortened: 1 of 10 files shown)


NuKeeper.Git/LibGit2SharpDriver.cs
```diff
@@ -84,7 +84,8 @@ public void CheckoutNewBranch(string branchName)
             var qualifiedBranchName = "origin/" + branchName;
             if (BranchExists(qualifiedBranchName))
             {
-                throw new NuKeeperException($"Git Cannot checkout new branch: a branch named '{qualifiedBranchName}' already exists");
+                _logger.Normal($"Git Cannot checkout new branch: a branch named '{qualifiedBranchName}' already exists");
```
MarcBruins: Wut? Why is this still a change..

NuKeeper.Inspection/NuGetApi/PackageVersionsLookup.cs
```diff
@@ -50,7 +50,7 @@ public PackageVersionsLookup(ILogger nuGetLogger, INuKeeperLogger nuKeeperLogger
             }
             catch (Exception ex)
             {
-                _nuKeeperLogger.Error($"Error getting {packageName} from {source}", ex);
+                _nuKeeperLogger.Normal($"Getting {packageName} from {source} returned exception: {ex.Message}");
```
MarcBruins: This is already in master...

MarcBruins:
This is ready to be reviewed

PR #654, opened 2019-01-19
Review use of data in SearchCodeRequest

AnthonySteele:
Review use of data in SearchCodeRequest
Make it read-only

### DIFF
--- a/NuKeeper.GitHub/OctokitClient.cs
+++ b/NuKeeper.GitHub/OctokitClient.cs
@@ -59,7 +59,12 @@ namespace NuKeeper.GitHub
             var user = await _client.User.Current();
             var userLogin = user?.Login;
             _logger.Detailed($"Read github user '{userLogin}'");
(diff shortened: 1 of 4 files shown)


NuKeeper.GitHub/OctokitClient.cs
```diff
@@ -59,7 +59,12 @@ public async Task<User> GetCurrentUser()
             var user = await _client.User.Current();
             var userLogin = user?.Login;
             _logger.Detailed($"Read github user '{userLogin}'");
-            return new User(user.Login, user.Name, user.Email);
+            if (user != null)
+            {
+                return new User(user.Login, user.Name, user.Email);
```
MarcBruins: can't you just do `return new User(user?.Login, user?.name, user?.Email)` and delete the if?

AnthonySteele: That would return a non-null user in every case

PR #657, opened 2019-01-20
Add Paket support

MarcBruins:
This PR adds paket support. It not complete yet and there a still issues.

The flow for a local `update` is as follows:

- [x] Read packages from `paket.dependencies` file in the `PaketRepositoryScanner`. This impl should be improved.
- [x] Do sorting, excluding, etc NuKeeper way
- [ ] Check for new version that satisfy the `paket.lock` file. Requires a Paket impl of `IPackageLookup` 
- [ ] Take those version and apply updates in the `PaketUpdateRunner`

For a remote repo there i still work to be done to read the `paket.dependecies` file.

### DIFF
--- a/NuKeeper.Inspection/RepositoryInspection/PaketRepositoryScanner.cs
+++ b/NuKeeper.Inspection/RepositoryInspection/PaketRepositoryScanner.cs
@@ -1,0 +1,49 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.FSharp.Core;
+using NuKeeper.Abstractions.Inspections.Files;
+using Paket;
+
+namespace NuKeeper.Inspection.RepositoryInspection
+{
+    public class PaketRepositoryScanner : IRepositoryScanner
+    {
+        private readonly IReadOnlyCollection<IPackageReferenceFinder> _finders;
+
+        public PaketRepositoryScanner(ProjectFileReader projectFileReader, PackagesFileReader packagesFileReader,
+            NuspecFileReader nuspecFileReader, DirectoryBuildTargetsReader directoryBuildTargetsReader)
+        {
+            _finders = new IPackageReferenceFinder[]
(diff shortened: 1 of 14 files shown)


NuKeeper/Engine/Packages/PackageUpdater.cs
```diff
@@ -74,7 +74,7 @@ public class PackageUpdater : IPackageUpdater
 
             foreach (var updateSet in updates)
             {
-                await _updateRunner.Update(updateSet, sources);
+                await _updateRunner.Update(updateSet, sources, null);
```
MarcBruins: We need to pass in a `IFolder` and not null so we can get the `paket.dependecies` later on.

NuKeeper.Inspection/RepositoryInspection/PaketRepositoryScanner.cs
```diff
@@ -0,0 +1,49 @@
...
+            _finders = new IPackageReferenceFinder[]
+                {projectFileReader, packagesFileReader, nuspecFileReader, directoryBuildTargetsReader};
+        }
+
+        public IReadOnlyCollection<PackageInProject> FindAllNuGetPackages(IFolder workingFolder)
+        {
+            var deps = new Dependencies($"{workingFolder.FullPath}/paket.dependencies");
+            try
+            {
+                var outdated = deps.FindOutdated(false, false, false, FSharpOption<string>.None); //todo see if we can get the current version list immeditaily
```
slang25: If we are looking to get all packages for NuKeeper to do the outdated logic, then we could use [GetDirectDependencies](https://github.com/fsprojects/Paket/blob/702432a917957e28ebb303541796c86f19e1ab46/src/Paket.Core/PublicAPI.fs#L563)

NuKeeper/Engine/Packages/PackageUpdater.cs
MarcBruins: Or some other way we can get access to the dependencies file of Paket

PR #658, opened 2019-01-22
Do not expose LibGit2Sharp usage outside of the NuKeeper.Git project

skolima:
Minor cleanup, might be useful if some replacement git driver is considered in order to support shallow checkout.


AnthonySteele:
I'm late to the party but 👍

PR #659, opened 2019-01-23
Trying out mixing command line git and LibGit2SharpDriver

skolima:


skolima:
Limitation: quite sure this will break (hang) when working with a git clone not created by NuKeeper and requiring password authentication. I'm open to suggestions how best to handle this - thinking about detecting this somehow and keeping `libgitsharp` in those cases.

skolima:
So, on any issues with the command, this hangs indefinitely. Not ideal.

MarcBruins:
Can we make this works for the local version first and then focus on the extra cloning steps? Would be very helpful to have this for a local `nukeeper repo {pat} ./` command. Why would it hang indefinitely? Could we read stdout and check for failures?

shep1987:
As #800 got merge can this be closed now?

skolima:
obsolete

PR #660, opened 2019-01-23
Don't supress exception

MarcBruins:
Collegeau of mine get's exception for the AzureDevops project but we don't what the exception is because it is not exposed.

### DIFF
--- a/NuKeeper.AzureDevOps/AzureDevOpsRestClient.cs
+++ b/NuKeeper.AzureDevOps/AzureDevOpsRestClient.cs
@@ -80,9 +80,9 @@ namespace NuKeeper.AzureDevOps
             }
             catch (Exception ex)
             {
-                msg = $"{caller}: {ex.Message}";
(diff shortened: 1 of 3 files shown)


NuKeeper.AzureDevOps/AzureDevOpsRestClient.cs
```diff
@@ -78,9 +78,9 @@ private async Task<T> HandleResponse<T>(HttpResponseMessage response, [CallerMem
             {
                 return JsonConvert.DeserializeObject<T>(responseBody);
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                msg = $"{caller}: Json exception - malformed PAT?";
+                msg = $"{caller}: {ex.Message}";
                 _logger.Error(msg);
                 throw new NuKeeperException(msg);
```
AnthonySteele: You can keep context on the excweption with `new NuKeeperException(msg, ex)`

Issue #661, opened 2019-01-24
NuKeeper fails to run on Azure DevOps when a repositories size is greater than int.MaxValue


sanderaernouts:
We have a few repositories for which we use [VFS for git](https://www.vfsforgit.org/). These repositories are >100 GB and do not contain code but other types of data.

When running NuKeeper against an Azure DevOps account that contains at least one git repository whose size is bigger than `int.MaxValue` the following exception occurs:
![image](https://user-images.githubusercontent.com/6184673/51658422-4c80ff80-1fa8-11e9-9bcc-57e095514cd7.png)

This is caused by a repository with a size of `51630689921`:
![image](https://user-images.githubusercontent.com/6184673/51658522-9a960300-1fa8-11e9-9350-e79d3e29103b.png)

I was able to fix it by changing the `Size` property of `AzureRepository` from `int` to `long`, I will submit a PR for this.

shep1987:
This has been release in 0.15.0 Let us know if you have any further issues :)

PR #662, opened 2019-01-24
Changed size property of AzureRepository from int to long.

sanderaernouts:
This PR fixes issue #661

# TL;DR
When you have a repository in your Azure DevOps project with a size greater than `int.MaxValue` you will get this error:
image

Updating from `int` to `long` fixes this issue. I tried to update to `ulong` but NewtonSoft JSON (de)serializes this as a long. Apparently, javascript does not support `ulong` values so JSON doesn't as a result of that. (https://stackoverflow.com/questions/29911786/json-net-prevent-ulong-from-getting-serialized-in-scientific-notation)


shep1987:
This has been release in 0.15.0 Let us know if you have any further issues :)

Issue #621, opened 2018-12-21
Azure DevOps 'Commit message has exceeded the maximum size.'.


MaxMommersteeg:
A pull request was created by NuKeeper (from Azure DevOps extension) and I am getting following message when clicking the `Set auto-complete` or `Complete` button on the pull request:
>Commit message has exceeded the maximum size.

![azuredevops_set_autocomplete](https://user-images.githubusercontent.com/9657173/50330740-a6e7f580-04fc-11e9-8c84-06c4f73b665c.png)

Following user voice could be related. However raising the character size will not always fix this issue. NuKeeper could potentially update thousands of packages in one go.
https://visualstudio.uservoice.com/forums/330519-azure-devops-formerly-visual-studio-team-services/suggestions/20217283-raise-the-character-limit-for-pull-request-descrip

Maybe we can limit character count in some way when dealing with Azure DevOps or have settings per platform? For now I will decrease the message size manually.

MaxMommersteeg:
Had a quick look at the source and come up with a small idea for resolving this issue.

Create an interface at `NuKeeper.Abstractions` called `ICommitWorder`. Have it specify the following signatures:
* `string MakePullRequestTitle(IReadOnlyCollection<PackageUpdateSet> updates);`
* `string MakeCommitMessage(PackageUpdateSet updates);`
* `string MakeCommitDetails(IReadOnlyCollection<PackageUpdateSet> updates);`

Rename `CommitWording` from `NuKeeper` to `DefaultCommitWording` and have it implemented `ICommitWorder`.

Add the following property to `CollaborationFactory`:
* `public ICommitWorder CommitWorder { get; private set; };`
Have it by default set to an instance of `DefaultCommitWording`. Default implementation could be overridden in `CollaborationFactory.CreateForPlatform` function.

We can then make other platform specific implementations of `ICommitWorder`. For example: `AzureDevOpsCommitWording`.

MarcBruins:
Will have a closer look soon, not now due to holidays

MarcBruins:
Could you make a start on this @MaxMommersteeg?

MaxMommersteeg:
Sure I'll try to have a concept PR open this weekend.

MaxMommersteeg:
A bit later than expected, but opened up an initial pull request ready for review.

MaxMommersteeg:
Fixed with https://github.com/NuKeeperDotNet/NuKeeper/pull/665.

PR #665, opened 2019-01-24
Platform specific commit wording.

MaxMommersteeg:
Do not merge this PR. It is currently work in progress.

This PR is my attempt to fix issue https://github.com/NuKeeperDotNet/NuKeeper/issues/621. I created the `ICommitWorder` interface that is added to the `CollaborationFactory` class as a property. The implementation of `ICommitWorder` should comply with the platform specific limitations given when opening up pull requests. For example: Pull request descriptions at Azure DevOps may not exceed the character count of 4000: https://visualstudio.uservoice.com/forums/330519-azure-devops-formerly-visual-studio-team-services/suggestions/20217283-raise-the-character-limit-for-pull-request-descrip

There are still a couple of things to do in order to fully implement this functionality, however I might need some guidance from the maintainers in order to continue:
- [x] Dealing with `PackageUpdateSet` in `NuKeeper.Abstractions`. `NuKeeper.Abstractions` needs the `PackageUpdateSet` from `NuKeeper.Inspections` but I don't think adding that dependency would be an ideal solution. Not sure what the correct option would be here.

- [x] Implement `AzureDevOpsCommitWorder` class. There are several possible solutions for the issue. We need to make a decision on how we should limit the total character count in the pull request description for Azure DevOps.
- [x] Write tests for `AzureDevOpsCommitWorder`. Should be straight forward when solution has been figured out.

For the time being I created an implementation for `ICommitWorder` called `DefaultCommitWorder`. This implementation behaves exactly as how it did before. Later on it is possible to create platform specific implementations for other platforms as well.


NuKeeper.Abstractions/CollaborationPlatform/ICommitWorder.cs
```diff
@@ -0,0 +1,13 @@
+using System.Collections.Generic;
+
+namespace NuKeeper.Abstractions.CollaborationPlatform
+{
+    public interface ICommitWorder
+    {
+        string MakePullRequestTitle(IReadOnlyCollection<PackageUpdateSet> updates);
```
MarcBruins: I don't think we should use `PackageUpdateSet` that is very specific for Nuget, if we are going to support `Paket` this is going to be a problem

MarcBruins: That seems like a good way forward.

MaxMommersteeg: Alright, I will try to get the concept finished this weekend.

MaxMommersteeg:
For now I cloned the `PackageUpdateSet` and related classes to NuKeeper.Abstractions. Next step is to strip these classes from excessive functionality. They should really only be simple POCO's.

Next is implementing the AzureDevOps specific CommitWorder to comply with Auzure DevOps's pull request message limitations.

Last step is to fix the unittests.

MaxMommersteeg:
@MarcBruins Could you have a quick look at `AzureDevOpsCommitWorder.cs`?

I removed `<details></details>` and `<summary></summary>` from the pull request body for Azure DevOps since it does not render (see earlier comment).

The way I prevent the pull request body from exceeding Azure DevOps its limits has big impact. This should be discussed on a functional level, not technical.

The current manner updates are shown in pull request bodies are too busy imho, example:
image
Because of that I made some changes on how updates are shown, using markdown tables (since markdown is supported by Azure DevOps). But perhaps this should be discussed in a different issue. 

I would really appreciate feedback from you and others before I proceed.

MarcBruins:
I'll have a look tomorrow Max, great that this is picked up again!

MaxMommersteeg:
Thanks for your feedback on this PR!

> Overall looks good, we are in the right direction. I'm curious to know why you moved so much classes to Abstractions? Wasn't the Inspection project enough?

A lot of classes from the `NuKeeper.Inspection` project were cloned since both CommitWorders (DefaultCommitWorder and AzureDevOpsCommitWorder) depend on those classes. Of course it is possible to remove those classes and add more properties to for example `PackageUpdateSet`.

I'm not really sure what you mean by `Wasn't the Inspection project enough?`. I can't have a reference from the `NuKeeper.Abstractions` project to the `NuKeeper.Inspection` project since that would result in a circular reference between the two.

NuKeeper.Abstractions/PackageInProject.cs
```diff
@@ -0,0 +1,30 @@
+using System.Collections.Generic;
+using System.Linq;
+using NuGet.Packaging.Core;
+using NuGet.Versioning;
+using NuKeeper.Abstractions.NuGet;
+
+namespace NuKeeper.Abstractions
+{
+    public class PackageInProject
```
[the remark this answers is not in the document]
MaxMommersteeg: `DefaultCommitWorder` and `AzureDevOpsCommitWorder` depend on its `Path` and `Version` properties.

MaxMommersteeg:
I made some more changes and got it to build. It currently only fixes https://github.com/NuKeeperDotNet/NuKeeper/issues/621. Issue https://github.com/NuKeeperDotNet/NuKeeper/issues/709 can be fixed once this pr is through.

PR #668, opened 2019-01-27
Short names must be one char. Pick new ones.

AnthonySteele:
With the latest `CommandlineUtils`, option short names must be one char (as intended). The 2-char short 
names don't work any more. 

This is a breaking change, but it is already broken.

In this PR, each option has a unique short name. Since there are 22 of them,  many are unrelated to the command text.
And we are running out of letters. Some future options won't have short names.

In this new scheme, some of the options don't have short names at all where they are 1) not frequently used and 2) no obvious letter is available.

### DIFF
--- a/NuKeeper.Abstractions.Tests/Configuration/FileSettingsReaderTests.cs
+++ b/NuKeeper.Abstractions.Tests/Configuration/FileSettingsReaderTests.cs
@@ -28,7 +28,7 @@ namespace NuKeeper.Abstractions.Tests.Configuration
             Assert.That(data.Include, Is.Null);
             Assert.That(data.Exclude, Is.Null);
             Assert.That(data.Label, Is.Null);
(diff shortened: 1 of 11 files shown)


AnthonySteele:
Now is a good time to revisit `MaxPr` / `MaxPullRequestsPerRepository` .

internally this is called `MaxPackageUpdates` as this more accurately reflects what it now does. 
If you're running in `repo` mode with one PR per package updated, then it is the number of PRs. Otherwise it's just a cap on the number of package updates.

This has caused confusion, and renaming has been suggested previously. Now is a good time to change this. Maybe to `MaxUpdates`  or `MaxPackageUpdates` ?

skolima:
Seems like `MaxPackageUpdates` is the name to use then. Perhaps something to suggest it's a per-repository limit?

PR #669, opened 2019-02-01
Github uri formats can vary

AnthonySteele:
Github uri formats can vary. not just `.git` on the end but also trailing slash.
This issue that prevented a fork being used due to presence/absence of trailing slash and therefor string mismatch.

Extracted the code that normalises them into `GithubHelpers`.  Call it in 1 more place where the format can vary.
Use `Uri` not `string` a bit more. 
More details in the message when it does not match. The original message doesn't say which of the 2 possible reasons it failed for, not pushable, or not matching. it just said
`User 'anthony-steele' fork of 'helper-lib' exists but is unsuitable.`

### DIFF
--- a/NuKeeper.Abstractions/Formats/UriFormats.cs
+++ b/NuKeeper.Abstractions/Formats/UriFormats.cs
@@ -1,6 +1,5 @@
 using System;
 using System.IO;
-using NuGet.Common;
 using NuKeeper.Abstractions.Git;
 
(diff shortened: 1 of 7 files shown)


AnthonySteele:
This issue was _probably_ introduced at #637

AnthonySteele:
Needs test with variations on urls in `GitHubForkFinderTests`

PR #678, opened 2019-02-14
fix(AzD): branch name comparison didn't work with AzD naming

### LINKED ISSUE #677: Crash with azuredevops extension
When running the azuredevops extension against my test-repo i get the following error:

```
Git Cannot checkout new branch: a branch named 'origin/nukeeper-update-Microsoft.NET.Test.Sdk-to-15.9.0' already exists
Updating'Microsoft.NET.Test.Sdk' from 15.8.0 to 15.9.0 in 1 projects
Selected package update of xunit.runner.visualstudio to 2.4.1
Selected package update of xunit to 2.4.1
Selected package update of Microsoft.NET.Test.Sdk to 15.9.0
Updates failed ArgumentNullException : Value cannot be null.
Parameter name: branch
Attempted 3 updates and did 0
```

Note that the branch doesn't exist, atleast not i can see it in the azuredevops dashboard. We will have to investigate further what is going on. Seems like the branch checkout fail causes this.

OneCyrus:
fixes #677

### DIFF
--- a/NuKeeper.Git/LibGit2SharpDriver.cs
+++ b/NuKeeper.Git/LibGit2SharpDriver.cs
@@ -146,9 +146,9 @@ namespace NuKeeper.Git
             {
                 
                 var localBranch = repo.Branches
-                    .FirstOrDefault(b => b.CanonicalName.EndsWith(branchName, StringComparison.OrdinalIgnoreCase));
+                    .Single(b => b.CanonicalName.EndsWith(branchName, StringComparison.OrdinalIgnoreCase));
                 var remote = repo.Network.Remotes
-                    .FirstOrDefault(b => b.Name.EndsWith(remoteName, StringComparison.OrdinalIgnoreCase));
+                    .Single(b => b.Name.EndsWith(remoteName, StringComparison.OrdinalIgnoreCase));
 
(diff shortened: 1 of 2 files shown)


NuKeeper.Git/LibGit2SharpDriver.cs
```diff
@@ -144,8 +144,11 @@ public void Push(string remoteName, string branchName)
 
             using (var repo = MakeRepo())
             {
-                var localBranch = repo.Branches[branchName];
-                var remote = repo.Network.Remotes[remoteName];
+                
+                var localBranch = repo.Branches
+                    .FirstOrDefault(b => b.CanonicalName.EndsWith(branchName, StringComparison.OrdinalIgnoreCase));
```
MarcBruins: This can `Single(b =>`

MarcBruins: Same here, we expect but one result and if not it should throw

OneCyrus: updated

OneCyrus: updated

PR #685, opened 2019-03-05
use dotnet format

AnthonySteele:
use dotnet format
A trial run of this tool:
https://github.com/dotnet/format
https://www.hanselman.com/blog/EditorConfigCodeFormattingFromTheCommandLineWithNETCoresDotnetFormatGlobalTool.aspx

### DIFF
--- a/NuKeeper.Abstractions.Tests/Configuration/FileSettingsReaderTests.cs
+++ b/NuKeeper.Abstractions.Tests/Configuration/FileSettingsReaderTests.cs
@@ -47,7 +47,7 @@ namespace NuKeeper.Abstractions.Tests.Configuration
         {
             var path = MakeTestFile("{}");
 
-            var fsr =  new FileSettingsReader(Substitute.For<INuKeeperLogger>());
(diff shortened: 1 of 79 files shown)


AnthonySteele:
It picked up a lot of minor spacing issues and inconsistencies that I had missed, and I don't yet see anything that's worse.
If you were to run it regularly, the diffs would be much smaller.
Looks good to me,

skolima:
Personally I've used StyleCopAnalyzers + settings to the same effect. Is it possible to plug in this tool into the build, so that it formats by default?

Side note, but related: the files without `\n` before EOF suggest the `.editorconfig` for NuKeeper is incomplete. I could tailor the one I use to fit the style in NuKeeper, if you're interested?

AnthonySteele:
I'm not familiar with getting StyleCop set up, I might take a look at it some time soon.
In the meantime, PRs are gladly accepted to enable StyleCop or to extend the `.editorconfig` here.

`dotnet format` is a good tool, [I had successes running it today](https://github.com/justeat/JustEat.StatsD/pull/169/files), so will use it again in some form, hopefully continuously or in smaller increments than this large PR!

Issue #682, opened 2019-02-25
A compatible SDK version for global.json version: [2.1.500]  was not found


AnthonySteele:
I get this error from one of our runs using the docker file:

```
Updating'Amazon.Lambda.Serialization.Json' from 1.4.0 to 1.5.0 in 1 projects
Updates failed NuKeeperException : Command dotnet failed with exit code: 145
A compatible SDK version for global.json version: [2.1.500] from [/tmp/NuKeeper/5d53bd1f7e7a4d0096fbaef4be0a5513/global.json] was not found
```

As far as I can see, what happens is:

The docker file has the .NET Core SDK version 2.2.x (and only that one) and NuKeeper . NuKeeeper runs fine, it pulls the code for this repo, finds a package update and then runs `dotnet` to perform the update. 

But this project has a `global.json` file containing 

```json
{
  "sdk": {
      "version": "2.1.500"
    }
}
```

so the `dotnet` command line to do the package update wants to use that specific SDK version, can't find it, and fails.

This project contains AWS lamdba functions, so it's going to stay on SDK `2.1.x` for a while (SDK `2.2.x` is not yet supported there).

I'm not sure how to fix this - should we have _two_ docker files, for SDK 2.1 and SDK 2.2 ?

AnthonySteele:
Actually, I'm thinking that we should soon have _three_ docker images, for SDK 2.1, SDK 2.2 and SDK 3.0-preview.

It's either that or have one big docker image, with multiple .NET SDKs installed on it.

slang25:
Given NuKeepers job is to update packages, I think it doesn't need to pay respect to the `global.json` SDK version, IMO it should use the latest stable SDK and ignore it.

Would running the commands from a different working directory get around it? If not, have it delete the file (but obviously not include that in the commits).

Am I thinking about this wrong?

AnthonySteele:
Right now I am looking at how other projects handle the "multiple base image" case.
We will likely want this to also target .NET Core `3.0 preview` alongside `2.x`.
What it looks like is that there is a folder hierarchy, with at most 1 `Dockerfile` per folder

e.g.
https://github.com/dotnet/dotnet-docker/tree/master/2.2/sdk/alpine3.8/amd64
https://github.com/nodejs/docker-node/tree/master/10/alpine

AnthonySteele:
I haven't had an issue when the `global.json` targets a slightly different version of .NET Core 2.1 to the installed SDK. 

[...]

PR #688, opened 2019-03-13
Two dockerfiles

AnthonySteele:
Two dockerfiles, one for each of the .NET Core SDK 2.1 and SDK 2.2

A fix /workaround for https://github.com/NuKeeperDotNet/NuKeeper/issues/682
This will require matching changes to builds on dockehub at https://cloud.docker.com/u/nukeeper/repository/docker/nukeeper/nukeeper/build
to build them both


MarcBruins:
How are you going to tag this in docker hub? I'm willing to try to see if this is a good solution but not yet convinced(but also no time to look into the issue)

PR #692, opened 2019-03-15
Try a larger "unit" of testing

AnthonySteele:
A proposed approach to tests, so supplement or partly replace current unit tests that focus on 1 class at a time ( As per [Mr Cooper's advice to avoid these highly coupled tests with lots of mocks](https://www.youtube.com/watch?v=EZ05e7EMOLM) ) and so make refactoring easier and tests simpler, and generally increase test coverage and therefor reliability.

Try an "end to end" find updates tests, with a sample `.csproj`  file as data, and a "find updates run against this test data. This exercises `UpdateFinder` and all code underneath it.  `UpdateFinder` is essentialy the entry point to `NuKeeper.Inspection`.

Anything under `NuKeeper.EndToEndTests/Data/` is not part of NuKeeper code, it is test data input to NuKeeper.

The idea is to extend this to a suite of sample projects, e.g. .NET core and .NET full framework, `.vbproj`, `.fsproj`, simple and quirky, and run similar "inspect" or even update tests against each of them.

The choice of  the package `"Newtonsoft.Json" Version="10.0.1" ` as starting point is deliberate - it's well known, and has a patch update to `10.03` as well as major updates.

Since these tests use a bare minimum of mocks, and real file and the real public nuget feed, you might call them "integration" tests, I don't really mind. It tests the behaviour of `NuKeeper.Inspection` via the entry point, not the implementation.

Related, some test coverage tracking would be good.

### DIFF
--- a/NuKeeper.EndToEndTests/Data/NewStyleCsProj/TestCase.csproj
+++ b/NuKeeper.EndToEndTests/Data/NewStyleCsProj/TestCase.csproj
@@ -1,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.1" />
+  </ItemGroup>
+</Project>
--- a/NuKeeper.EndToEndTests/FinderTests.cs
+++ b/NuKeeper.EndToEndTests/FinderTests.cs
@@ -1,0 +1,81 @@
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using NuKeeper.Abstractions.Configuration;
+using NuKeeper.Abstractions.Logging;
+using NuKeeper.Abstractions.NuGet;
+using NuKeeper.Inspection;
+using NuKeeper.Inspection.Files;
+using NuKeeper.Inspection.NuGetApi;
+using NuKeeper.Inspection.RepositoryInspection;
+using NUnit.Framework;
+
+namespace NuKeeper.EndToEndTests
+{
+    public class FinderTests
+    {
+        [Test]
+        public async Task TestMissingFolder()
+        {
+            var logger = new TestLogger();
+            var finder = BuildUpdateFinder(logger);
+
+            Assert.That(finder, Is.Not.Null);
+
+            var updates = await finder.FindPackageUpdateSets(
+                FolderForPath(".{sep}NoSuchPath"),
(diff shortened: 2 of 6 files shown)


skolima:
Hmm. My instinct here would be to use System.IO.Abstractions or at least a temp folder for the tests. This somehow feels... wrong (with the caveat that a lot of what Ian says in the talk is against how TDD tends to be practiced - that's what makes it a great talk). Thoughts?

AnthonySteele:
So, I will have a look at [the file system abstractions](https://github.com/System-IO-Abstractions/System.IO.Abstractions) - I think that it might be useful.

I'm hoping to get a decent body of tests, enough that managing them with real files and folders will be _much_ easier than literal strings in test code.  Could that file system abstraction fake part of the path, e.g. "this folder is `C:\Code\NuKeeper\NuKeeper.EndToEndTests\bin\Debug\netcoreapp2.1\Data`, but pretend that it is `C:\Data` "? Being able to do this would also mean that the `IDirectoryExclusions` would not have to be mocked, which is a good sign.

I think that some of Ian's points are very applicable - we have tests "per class" with lots of mocks, and this makes refactoring these classes hard, tests are verbose and hard to follow, etc. If tests can be done at the entry point to `NuKeeper.Inspection` with coverage tracking, then maybe this could result in both higher test coverage (keeping some tests on subsystems where needed) and  code that is easier to work with. e.g. that `BuildUpdateFinder` method necessary in the tests is a place to start looking at complexity.

skolima:
System.IO.Abstractions replace your file system access with a set of fakes (in test mode). Mixing actual file system with fakes in tests is probably ill-advised. However, your exclusion etc. would also all work on the fake filesystem, meaning, you could have any paths you like.

AnthonySteele:
I'm trying to picture how it would work when there is a test suite of e.g. 10 or 20 input cases, each of which is in theory a working project or solution, i.e. can be a  `.sln` file, subfolders and multiple `.csproj` (  or `vbproj` ) files plus `packages.config` files.

managing that on the file system seems much easier than managing that as strings embedded in c# code.

skolima:
That's not a supported scenario for S.IO.A, unfortunately.

AnthonySteele:
Actually, making a copy under a `temp` folder, running tests and then deleting that has advantages: it's easier to do tests that actually perform updates, i.e. alter the files.

slang25:
I had a play about a few months ago with using MSBuild to do the processing of proj files, and Paket. Both of these want actual files, I think it's also more manageable fwiw.

AnthonySteele:
Right, if a test encompasses `dotnet nuget update` or the like, it will have to work on a real file system.

AnthonySteele:
Will brig this back in another form, the concept is IMHO a good one. The implementation needs some work.

Issue #327, opened 2018-07-12
Gitlab Integration


reaction1989:
ATM there is only a GitHub Integration. 

Having an GitLab MR Integration would open the door for many more users

AnthonySteele:
This would be a useful feature. The hard part would probably be in the project organisation and appropriate interfaces and DTOs to handle either GitHub or GitLab or other integrations.

SeriousM:
May I add the request to support VSTS GIT Repositories?
This would be awesome!!

AnthonySteele:
If they have an API that we can work to do equivalent things to the github api, then this is all possible.

lxalln:
There is a full API for GitLab.

[GitLab Merge Request API docs](https://docs.gitlab.com/ee/api/merge_requests.html)

phatcher:
@AnthonySteele Is it worthwhile to try and spec this out or should I just dive in and attempt something?

The best approach I can see is to work on VSTS/GitLab specific APIs and then lift out commonality once each of these is working.  Advantage of this is we don't disrupt the development/maintenance of the main code line whilst this is going on.

slang25:
@AnthonySteele It might be worth adopting an "up for grabs" label for these issues

AnthonySteele:
True, I won't be able to work on VSTS integration since I have no VSTS to integrate with, so it is "up for grabs". 

I expected the first step is to introduce some abstractions that could be implemented by github and also by VSTS, gitlab etc and so decouple a lot of the code from the specifics of the guthub impl 

If it's possible, i would suggest trying to make a "VSTS Integration" project, trying to get some commonality with github.

AnthonySteele:
Similar to #336 - supporting one of VSTS / Gitlab should help the other

MarcBruins:
We have support now for VSTS, BitBucket and  Github, adding this shouldn't be hard.

zubivan:
Hi all, I picked it up. Will create a pull request with initial structure.

zubivan:
Looks like we can close this one? @MarcBruins @AnthonySteele

MarcBruins:
You are right @zubivan, thanks for the notice!

PR #696, opened 2019-03-28
GitLab support

zubivan:
Started to work on GitLab (issue #327) support for NuKeeper based on https://docs.gitlab.com/ee/api/

What's already there:
- Implementation of settings reader + some barebones tests
- Basic implementation of gitlab rest client and token authentication
- `GetCurrentUser` api call

Looking for some initial feedback.

I see quite some duplication in client implementations for Azure/Bitbucket and co. I thought first to focus on bringing Single repository support for GitLab and continuing with a cleanup a bit later. What do you guys think?

Update:
- Finished the implementation for a single repository

TODO:
- Unit tests

### DIFF
--- a/NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
+++ b/NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
@@ -9,13 +9,11 @@ namespace NuKeeper.Gitlab.Tests
     {
         private GitlabSettingsReader _gitlabSettingsReader;
 
-
         [SetUp]
         public void Setup()
         {
             _gitlabSettingsReader = new GitlabSettingsReader();
         }
-
 
         [Test]
         public void ReturnsCorrectPlatform()
@@ -24,7 +22,6 @@ namespace NuKeeper.Gitlab.Tests
 
             Assert.AreEqual(Platform.GitLab, platform);
         }
-
 
         [Test]
         public void AssumesItCanReadGitLabUrls()
--- a/NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
+++ b/NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
@@ -41,7 +41,7 @@ namespace NuKeeper.Gitlab.Tests
             var repositorySettings = _gitlabSettingsReader.RepositorySettings(repositoryUri, "master");
 
             Assert.IsNotNull(repositorySettings);
-            Assert.AreEqual(new Uri("https://gitlab.com/api/v4"), repositorySettings.ApiUri);
+            Assert.AreEqual(new Uri("https://gitlab.com/api/v4/"), repositorySettings.ApiUri);
             Assert.AreEqual(repositoryUri, repositorySettings.RepositoryUri);
             Assert.AreEqual("user", repositorySettings.RepositoryOwner);
             Assert.AreEqual("projectname", repositorySettings.RepositoryName);
--- a/NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
+++ b/NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
@@ -41,7 +41,7 @@ namespace NuKeeper.Gitlab.Tests
             var repositorySettings = _gitlabSettingsReader.RepositorySettings(repositoryUri, "master");
 
             Assert.IsNotNull(repositorySettings);
-            Assert.AreEqual(new Uri("https://gitlab.com/api/v4/"), repositorySettings.ApiUri);
+            Assert.AreEqual(new Uri("https://gitlab.com/api/v4"), repositorySettings.ApiUri);
             Assert.AreEqual(repositoryUri, repositorySettings.RepositoryUri);
             Assert.AreEqual("user", repositorySettings.RepositoryOwner);
             Assert.AreEqual("projectname", repositorySettings.RepositoryName);
--- a/NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
+++ b/NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
@@ -1,5 +1,7 @@
 using System;
+using NSubstitute;
 using NuKeeper.Abstractions.CollaborationPlatform;
+using NuKeeper.Abstractions.Configuration;
 using NUnit.Framework;
 
 namespace NuKeeper.Gitlab.Tests
@@ -8,11 +10,14 @@ namespace NuKeeper.Gitlab.Tests
     public class GitlabSettingsReaderTests
     {
         private GitlabSettingsReader _gitlabSettingsReader;
+        private IEnvironmentVariablesProvider _environmentVariablesProvider;
 
         [SetUp]
         public void Setup()
         {
-            _gitlabSettingsReader = new GitlabSettingsReader();
+            _environmentVariablesProvider = Substitute.For<IEnvironmentVariablesProvider>();
+
+            _gitlabSettingsReader = new GitlabSettingsReader(_environmentVariablesProvider);
         }
 
         [Test]
@@ -21,6 +26,21 @@ namespace NuKeeper.Gitlab.Tests
             var platform = _gitlabSettingsReader.Platform;
 
             Assert.AreEqual(Platform.GitLab, platform);
+        }
+
+        [Test]
+        public void UpdatesAuthenticationTokenFromTheEnvironment()
+        {
+            _environmentVariablesProvider.GetEnvironmentVariable("NuKeeper_gitlab_token").Returns("envToken");
+
+            var settings = new CollaborationPlatformSettings
+            {
+                Token = "accessToken",
+            };
+
+            _gitlabSettingsReader.UpdateCollaborationPlatformSettings(settings);
+
+            Assert.AreEqual(settings.Token, "envToken");
         }
 
         [Test]
--- a/NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
+++ b/NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
@@ -31,18 +31,19 @@ namespace NuKeeper.Gitlab.Tests
             Assert.AreEqual(true, canRead);
         }
 
-        [Test]
-        public void GetsCorrectSettingsFromTheUrl()
+        [TestCase(null)]
+        [TestCase("master")]
+        public void GetsCorrectSettingsFromTheUrl(string targetBranch)
         {
             var repositoryUri = new Uri("https://gitlab.com/user/projectname.git");
-            var repositorySettings = _gitlabSettingsReader.RepositorySettings(repositoryUri, "master");
+            var repositorySettings = _gitlabSettingsReader.RepositorySettings(repositoryUri, targetBranch);
 
             Assert.IsNotNull(repositorySettings);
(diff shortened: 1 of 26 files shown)


zubivan:
image

Alright, almost there!
Doing some cleanup and the first semi functional version should be reviewable.

MarcBruins:
Great stuff! We are getting there! Had some comments on the code and still some a lot spot that are not implemented yet. If you have questions are want some feedback we are also on gitter: https://gitter.im/NuKeeperDotNet/Lobby

NuKeeper.Gitlab/GitlabPlatform.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        public Task<IReadOnlyList<Organization>> GetOrganizations()
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task<IReadOnlyList<Repository>> GetRepositoriesForOrganisation(string projectName)
+        {
+            throw new NotImplementedException();
```
MarcBruins: Delete it if we are not using it

NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
```diff
@@ -0,0 +1,51 @@
...
+using NUnit.Framework;
+
+namespace NuKeeper.Gitlab.Tests
+{
+    [TestFixture]
+    public class GitlabSettingsReaderTests
+    {
+        private GitlabSettingsReader _gitlabSettingsReader;
+
+
```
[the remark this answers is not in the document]
zubivan: Should run dotnet format against the whole thing...
Trying out Rider on Mac while building it, so might have missed formatting in some places. Will fix.

NuKeeper.Gitlab/GitlabPlatform.cs
```diff
@@ -0,0 +1,97 @@
...
+            var mergeRequest = new MergeRequest
+            {
+                Title = request.Title,
+                SourceBranch = request.Head,
+                Description = request.Body,
+                TargetBranch = request.BaseRef,
+                Id = $"{projectName}/{repositoryName}"
+            };
+
+            await _client.OpenMergeRequest(projectName, repositoryName, mergeRequest).ConfigureAwait(false);
```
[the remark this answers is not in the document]
zubivan: I do agree here, however gitlab is actually calling it Merge Request instead of Pull Request, so it might be cleaner to keep it this way here as it's anyway a platfrom specific code.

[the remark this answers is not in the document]
zubivan: Obviously your call :)

NuKeeper.Gitlab/GitlabSettingsReader.cs
```diff
@@ -0,0 +1,69 @@
...
+    public class GitlabSettingsReader : ISettingsReader
+    {
+        private const string GitLabTokenEnvironmentVariableName = "NuKeeper_gitlab_token";
+        private const string UrlPattern = "https://gitlab.com/{username}/{projectname}.git";
+
+        public Platform Platform => Platform.GitLab;
+
+        public bool CanRead(Uri repositoryUri)
+        {
+            return repositoryUri?.Host.Contains("gitlab", StringComparison.OrdinalIgnoreCase) == true;
```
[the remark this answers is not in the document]
zubivan: Oh, yeah. Will fix.

NuKeeper.Gitlab/GitlabPlatform.cs
zubivan: Part of the interface implementation, so can't just remove it. One thing I can do here is to add error logging to make it more explicit that this functionality is not implemented on purpose.

NuKeeper.Gitlab/GitlabPlatform.cs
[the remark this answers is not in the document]
MarcBruins: So `Azure DevOps` uses `CreatePullRequest` and Github uses `OpenPullRequest`..

NuKeeper.Gitlab/GitlabPlatform.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        public Task<IReadOnlyList<Organization>> GetOrganizations()
+        {
+            throw new NotImplementedException();
```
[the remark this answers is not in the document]
zubivan: There is a similar concept to organisations (projects). I can probably pick it up with one of the next PRs.

NuKeeper.Gitlab/GitlabSettingsReader.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public RepositorySettings RepositorySettings(Uri repositoryUri, string targetBranch = null)
+        {
+            if (repositoryUri == null)
+            {
+                throw new NuKeeperException(
+                    $"The provided uri was is not in the correct format. Provided null and format should be {UrlPattern}");
+            }
+
+            // Assumption - url should look https://gitlab.com/{username}/{projectname}.git";
```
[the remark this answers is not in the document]
zubivan: Here we can't really be sure, because GitLab can also be hosted on premises. But we have some structure checks following later. Taking similar approach as in Bitbucket or Azure implementation.

NuKeeper.Gitlab.Tests/GitlabSettingsReaderTests.cs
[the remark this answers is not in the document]
zubivan: Fixed

NuKeeper.Gitlab/GitlabPlatform.cs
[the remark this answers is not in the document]
MarcBruins: Keeping platform standards is also fine

NuKeeper.Gitlab/GitlabPlatform.cs
MarcBruins: That would be a nice

zubivan: @MarcBruins 👍already pushed those changes

NuKeeper.Gitlab.Tests/NuKeeper.Gitlab.Tests.csproj
```diff
@@ -0,0 +1,20 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="nunit" Version="3.10.1" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.10.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
```
slang25: `16.0.1`

NuKeeper.Gitlab.Tests/NuKeeper.Gitlab.Tests.csproj
```diff
@@ -0,0 +1,20 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="nunit" Version="3.10.1" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.10.0" />
```
slang25: `3.13.0`

NuKeeper.Gitlab.Tests/NuKeeper.Gitlab.Tests.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="nunit" Version="3.10.1" />
```
slang25: `3.11.0`, and `nunit`->`NUnit`

NuKeeper.Gitlab/GitlabRestClient.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            if (!response.IsSuccessStatusCode)
+            {
+                _logger.Detailed($"Response {response.StatusCode} is not success, body:\n{responseBody}");
+
+                if (customErrorHandling != null)
+                {
+                    var result = customErrorHandling(response.StatusCode);
+                    if (result.IsSuccessful)
+                        return result.Value;
```
slang25: I love F# and `Result` type, but I'm finding the usage here a little odd. Seemingly this is the only place where it's unpacked, there's only one call to here where `customErrorHandling` is not `null`, the name `customErrorHandling` seems off as it's mapping a status code to a result, and the `Result` type should instead have a `Match` method take takes success and failure `Func`s IMO.

AnthonySteele: Interesting. I have used result wrapper types many times, idea taken from too many places to mention.  It's definitely applicable outside of just GitLab integration, and can have more functional extensions, but that can come in later refactoring.

zubivan: What would be your suggestion here, guys?
Currently it's here just to handle 404 coming from the API. That's a valid response for non existing branch.

I can probably move the `Result` type to a `Nukeeper.Abstractons` assembly and make it more consistent with i.e. F# implementation.

NuKeeper.Gitlab.Tests/NuKeeper.Gitlab.Tests.csproj
zubivan: Done

NuKeeper.Gitlab.Tests/NuKeeper.Gitlab.Tests.csproj
zubivan: Done

NuKeeper.Gitlab.Tests/NuKeeper.Gitlab.Tests.csproj
zubivan: Done

NuKeeper.Gitlab/GitlabSettingsReader.cs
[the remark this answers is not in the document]
zubivan: That should be fixed

NuKeeper.Gitlab/GitlabRestClient.cs
slang25: On further reflection, what is there currently is fine, it's a private class so it can always be changed.

It doesn't have to look like the F# implementation, but I found myself unfamiliar with what I expected (I'm easily confused). However, looking at the one place it's called it makes no sense to add a `.Match(Action<T> onSuccess, Action onFailure)` or `.WithDefault(T defaultValue)`, because we want to branch and conditionally return anyway.

My only last thought is that it might be better suited to a tuple return `(bool isSuccessful, T result)`, but I'm just nit-picking 🙂

zubivan: Thinking it might be a bit confusing to get a tuple containing both `(bool isSuccessful, T result)`, since result only makes sense when `isSuccessful` is true.
With current implementation trying to get the `Value` will throw an exception at you when `IsSuccessful` is false. 🤔 

Let me know what you think. 🙂

slang25: It's the problem as the Result type here, except it throws a more explicit exception. Ignore me, I'm not progressing anything here 🙂

PR #697, opened 2019-03-29
Simplifies the targetbranch implementation and also fixes a bug in wh…

MarcBruins:
Fixes a bug and simplifies the targetbranch implementation, it now relies on RemoteInfo.Branch.

### DIFF
--- a/NuKeeper.Abstractions/CollaborationModels/PullRequestRequest.cs
+++ b/NuKeeper.Abstractions/CollaborationModels/PullRequestRequest.cs
@@ -6,12 +6,12 @@ namespace NuKeeper.Abstractions.CollaborationModels
         {
             Head = head;
             Title = title;
-            BaseRef = baseRef;
+            BaseRef = baseRef.Replace("origin/", string.Empty);
         }
 
         public string Head { get; }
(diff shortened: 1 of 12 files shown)


NuKeeper.Abstractions/CollaborationModels/PullRequestRequest.cs
```diff
@@ -6,12 +6,14 @@ public PullRequestRequest(string head, string title, string baseRef)
         {
             Head = head;
             Title = title;
-            BaseRef = baseRef;
+
+            //This can be a remote that has been passed in, this happens when run locally against a targetbranch that is remote
```
slang25: Nit: just call Replace, the check is redundant

MarcBruins: That's what I had before, but then every time a unnecessary replace is called.

slang25: I think that makes sense though: https://stackoverflow.com/questions/34404042/should-i-use-string-contains-before-string-replace

PR #701, opened 2019-04-04
Fix regression in 0.19.1

skolima:
#697 broke support for cloning remote git repositories


MarcBruins:
That's bad. We should have had some tests failures there..

MarcBruins:
I'll push a 0.19.2 asap

MarcBruins:
Should be available now

PR #702, opened 2019-04-05
Update the analysers

AnthonySteele:
Update the analysers.
Remove `Text.Analyzers` it seems to be deprecated.

The new version of analysers is a lot stricter on exception types, with  [CA1031: do not catch general exception types](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1031-do-not-catch-general-exception-types?view=vs-2019).  So general `catch` blocks have to be fixed, either
* Catch a more specific exception type - this is used where an exception type or base type can be determined. I have used
  * `IOException`: base class for file and directory exceptions.
  * `ApiException`: base class for octokit github exceptions.
  * `JsonException`: base class for Newtonsoft Json exceptions.
* Removed the block entirely - this is used where null checks in the code are the right way to do it
* Use a `#pragma` to ignore the warning for a line - this is used as a last resort, or where there is a lot of code / multiple possible providers inside the block and so an exception type is not simple to determine,.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -16,9 +16,8 @@
     <RepositoryUrl>https://github.com/NuKeeperDotNet/NuKeeper.git</RepositoryUrl>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.3" PrivateAssets="All" />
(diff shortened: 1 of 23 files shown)


MarcBruins:
Can you try a new release and make sure you add some text to the "release description" field? Should trigger a new github release with listed changes

Issue #710, opened 2019-04-12
Gitea Integration


sharpSteff:
ATM Gitea (a open source, self hosted Git service, https://try.gitea.io/) is not integrated.

For a lot of users, it's the best free (open source) alternative, when it comes to self hosted git services.

MaxMommersteeg:
Gitea integration has been implemented and this issue can be closed right?

PR #711, opened 2019-04-12
Gitea Integration

sharpSteff:
Started to work on Gitea(issue #710 ) support for NuKeeper based on https://try.gitea.io/api/swagger# and https://docs.gitea.io/en-us/

What's already there:

Implementation of settings reader + some tiny unit tests
Basic implementation of gitea rest client and token authentication including:

- GetCurrentUser
- GetRepository
- GetOrganizations (only for admins)
- GetRepositoryBranch
- ForkRepository
- OpenPullRequest

At the end a lot of code in the restclient is similar to GitHub/Gitlab. Maybe a common base class would remove this redundant code.

TODO:
- ~GiteaPlattform.GetRepositoriesForOrganisation (and REST)~
- GiteaPlattform.Search (and REST)
- ~Enhance GiteaRepositoryDiscovery.GetRepositories for ServerScope.Global and ServerScope.Organisation~
- ~GiteaForkFinder implement other functionality than ForkMode.SingleRepositoryOnly~

Unit tests
- Use real repository for unit tests (??)


sharpSteff:
pullrequest_nukeeper
gitea_pull

Here a some example screens, after forking nukeeper to a local hosted gitea and running nukeeper against it :)

AnthonySteele:
This is great!

Agreed that soon we should focus on common code across collaboration platforms (perhaps a `CollaborationPlatform.Abstractions` project?  ) And on getting them better tested.

MarcBruins:
Could you also add some docs on how to use it? Something like this https://github.com/NuKeeperDotNet/NuKeeper/pull/720/files#diff-32cb44efbeb7ab3bf9cf2edef28be295R69 Its just to show other people how it works. There is a `gitea.md` file available here: https://github.com/NuKeeperDotNet/NuKeeper/blob/master/site/content/platform/gitea.md. If you pull master you can edit that file and add it to this PR.

MarcBruins:
The bitbucket explanation is probably a better startpoint: https://github.com/NuKeeperDotNet/NuKeeper/pull/724/files#diff-dce3466c4d63c7bf1684bbfa5a2e235f

sharpSteff:
@MarcBruins Thanks for the startpoint!

Following points are also finishe:
- Enhance GiteaRepositoryDiscovery.GetRepositories for ServerScope.Global and ServerScope.Organisation
- GiteaForkFinder implement other functionality than ForkMode.SingleRepositoryOnly

so automatically forking from an existing repository also works from now on

sharpSteff:
@MarcBruins 
I recognized ForkMode has no default value in FileSettings.
Is this intended ?

AnthonySteele:
The default is (or should be) `PreferFork`

sharpSteff:
when there is no nukeeper.settings.json-config file (which i has to create manuelly, by the way) or a nukeeper.settings.json-config without **ForkMode** defined, **ForkMode** might be null since its explicitly declared as a nullable type in **FileSettings**

`public ForkMode? ForkMode { get; set; }`

In **CollaborationFactory.ValidateSettings()** ForkMode will be checked 

` if (!Settings.ForkMode.HasValue)
            {
                return ValidationResult.Failure("Fork Mode was not set");
            }`
In result, the operation will be cancelled.

However when a nukeeper.settings.json is available with ForkMode defined everything works as expected

NuKeeper.Gitea/GiteaPlatform.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        public async Task<IReadOnlyList<Organization>> GetOrganizations()
+        {
+            var orgas = await _client.GetOrganizations();
+            return orgas?.Select(x => MapOrganization(x)).ToList() ?? new List<Organization>();
+        }
+
+        public async Task<IReadOnlyList<Repository>> GetRepositoriesForOrganisation(string organizationName)
+        {
+            var repos = await _client.GetOrgaRepositories(organizationName);
```
MarcBruins: Type in `GetOrga`? Should be `GetOrganisationRepositories`?

NuKeeper.Gitea/GiteaRestClient.cs
```diff
@@ -0,0 +1,189 @@
...
+            _client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("token", token);
+        }
+
+        /// <summary>
+        /// GET /user https://try.gitea.io/api/swagger#/user/userGetCurrent
+        /// </summary>
+        /// <returns>returns the current user</returns>
+        public async Task<User> GetCurrentUser()
+        {
+            return await GetResource<User>("user");
```
MarcBruins: This is correct? Just put in "user" for a request gets you the current user?

NuKeeper.Gitea/GiteaSettingsReader.cs
```diff
@@ -0,0 +1,103 @@
...
+                // There is no real identifier for gitea repos so try to get the gitea swagger json
+                var client = new HttpClient()
+                {
+                    BaseAddress = new Uri(baseaddress)
+                };
+
+                client.DefaultRequestHeaders.Accept.Add(
+                    new MediaTypeWithQualityHeaderValue("application/json"));
+
+                HttpResponseMessage response = client.GetAsync($"swagger.v1.json").Result;
```
MarcBruins: Please no `.Result`. This will block the code untill it's complete

NuKeeper.Gitea/GiteaSettingsReader.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            return new RepositorySettings
+            {
+                ApiUri = uriBuilder.Uri,
+                RepositoryUri = repositoryUri,
+                RepositoryName = repoName,
+                RepositoryOwner = repoOwner,
+                RemoteInfo = targetBranch == null
+                    ? null
+                    : new RemoteInfo { BranchName = targetBranch }
```
MarcBruins: Wow, you also using targetbranch here. Is there business value for you in using that?

NuKeeper.Gitea/Model/UserPermissions.cs
```diff
@@ -0,0 +1,28 @@
+using Newtonsoft.Json;
+
+namespace NuKeeper.Gitea.Model
+{
+    //public class UserPermissions
```
MarcBruins: Please remove if not used

MarcBruins:
> 
> 
> when there is no nukeeper.settings.json-config file (which i has to create manuelly, by the way) or a nukeeper.settings.json-config without **ForkMode** defined, **ForkMode** might be null since its explicitly declared as a nullable type in **FileSettings**
> 
> `public ForkMode? ForkMode { get; set; }`
> 
> In **CollaborationFactory.ValidateSettings()** ForkMode will be checked
> 
> ` if (!Settings.ForkMode.HasValue) { return ValidationResult.Failure("Fork Mode was not set"); }`
> In result, the operation will be cancelled.
> 
> However when a nukeeper.settings.json is available with ForkMode defined everything works as expected

Can you make `preferFork` the default?

NuKeeper.Gitea/GiteaPlatform.cs
sharpSteff: should be
```suggestion
            var repos = await _client.GetOrganisationRepositories(organizationName);
```

NuKeeper.Gitea/GiteaRestClient.cs
sharpSteff: yes https://try.gitea.io/api/swagger#/user/userGetCurrent
quite easy isn't it ;)
 GET /user

NuKeeper.Gitea/GiteaSettingsReader.cs
sharpSteff: issue here was CanRead() is no Task<bool>. So I had to go the sync way
Any recommandations how to solve this ?

MarcBruins: So the only way to know if it's a gitea server is to call the swagger api? There is no convention whatsoever that must be used? If not, i think we should have an extra parameter which specifies that it's the desired platform: `--platform gitea`

MarcBruins: I think we discuss this approach earlier on gitter or somewhere else.. I don't want to make this a blocker, would be nice if we could add it but it doesn't have to be in this PR. I could be a new upforgrabs feature.

NuKeeper.Gitea/GiteaRestClient.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            return await GetResource<List<Organization>>("admin/orgs");
+        }
+
+        /// <summary>
+        /// /GET /orgs/{org}/repos List an organization's repos https://try.gitea.io/api/swagger#/organization/orgListRepos
+        /// </summary>
+        /// <param name="orgaName">name of the organization</param>
+        /// <returns>list of repos </returns>
+        public async Task<List<Repository>> GetOrgaRepositories(string orgaName)
```
MarcBruins: Please change to proper naming like discussed above

NuKeeper.Gitea/GiteaPlatform.cs
MarcBruins: Could you update that?

NuKeeper.Gitea/GiteaSettingsReader.cs
sharpSteff: `CanRead` - does only provide the repo-uri. Since gitea is most of the time self hosted, the repo uri looks like this
https://localhost:12345/{username}/{projectname}.git

So trying to get the json from https://yourgitea/swagger.v1.json (maybe even trying to validate this json, since there are gitea-version, etc stored) was somehow the safest way.

I think when nukeeper, wants to support self hostet github or vsts services, this issue will appear again sooner or later.

a desired platform-parameter would solve this issue however

AnthonySteele: We have the ` --platform` option already. See https://github.com/NuKeeperDotNet/NuKeeper/wiki/Configuration 

As the docs say:
> This is typicaly infered from the api url structure, but since this does not always work, it can be specified here if neccessary.

sharpSteff: @AnthonySteele so when using `--platform`, the [CanRead(uri) ](url) will be skipped

sharpSteff:
> Can you make `preferFork` the default?

I think I'll create an additioanl issue for this

PR #712, opened 2019-04-12
Add nukeeper website structure

MarcBruins:
This adds the nukeeper website structure.  It's a [Hugo website](https://gohugo.io/) which uses the [Learn theme](https://learn.netlify.com/en/).

It enables us to update the website as markdown files and have docs update with every breaking change/feature update that we have. The markdown is hosted beside the code in a `site` folder, every change to the .MD file regenerates the website.

I suggest we take this as a base and iterate upon it, at the time we are comfortable with the website and all Wiki content is moved we can close the wiki.


site/config.toml
```diff
@@ -0,0 +1,13 @@
+languageCode = "en-us"
+title = "NuKeeper"
+theme = "hugo-theme-learn"
+
```
slang25:
```suggestion
enableEmoji = true
```

slang25: This will show the emoji checkmarks where we use them

MarcBruins: wow we definitely need that!

MarcBruins: done 👍 ⚡️

PR #730, opened 2019-04-16
--targetBranch support for GitHub

Chama-StephanTuinder:
To make it possible to run NuKeeper on a different branch in GitHub

Case: 
 updating of the nugets in feature-branches of different projects that work on the same feature.
In this way the update of one projects nugets will be populated automaticly to the other projects wich work on the same feature.

### DIFF
--- a/NuKeeper.Abstractions/Git/IGitDriver.cs
+++ b/NuKeeper.Abstractions/Git/IGitDriver.cs
@@ -9,6 +9,8 @@ namespace NuKeeper.Abstractions.Git
 
         void Clone(Uri pullEndpoint);
 
+        void Clone(Uri pullEndpoint, string branchName);
+
         void AddRemote(string name, Uri endpoint);
 
         void Checkout(string branchName);
--- a/NuKeeper.GitHub/GitHubSettingsReader.cs
+++ b/NuKeeper.GitHub/GitHubSettingsReader.cs
@@ -59,7 +59,7 @@ namespace NuKeeper.GitHub
                 RepositoryUri = repositoryUri,
                 RepositoryName = repoName,
                 RepositoryOwner = repoOwner,
-                RemoteInfo = targetBranch != null ? new RemoteInfo { BranchName = $"origin/{targetBranch}" } : null
+                RemoteInfo = targetBranch != null ? new RemoteInfo { BranchName = targetBranch } : null
             };
         }
     }
--- a/NuKeeper.GitHub/GitHubSettingsReader.cs
+++ b/NuKeeper.GitHub/GitHubSettingsReader.cs
@@ -59,7 +59,7 @@ namespace NuKeeper.GitHub
                 RepositoryUri = repositoryUri,
                 RepositoryName = repoName,
                 RepositoryOwner = repoOwner,
-                RemoteInfo = ( targetBranch != null ? new RemoteInfo() { BranchName = $"origin/{targetBranch}" } : null )
+                RemoteInfo = targetBranch != null ? new RemoteInfo { BranchName = $"origin/{targetBranch}" } : null
             };
         }
     }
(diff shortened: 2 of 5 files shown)


NuKeeper/Engine/RepositoryUpdater.cs
```diff
@@ -108,6 +108,9 @@ private static void GitInit(IGitDriver git, RepositoryData repository)
         {
             git.Clone(repository.Pull.Uri);
             repository.DefaultBranch = repository.DefaultBranch ?? git.GetCurrentHead();
+            // checkout will change the branch to the correct one if it is changed
+            // with the --targetBranch option
+            git.Checkout(repository.DefaultBranch);
```
MarcBruins: What happens if we already checkout a branch? Is that Checkout done twice now?

NuKeeper.GitHub/GitHubSettingsReader.cs
```diff
@@ -59,6 +59,7 @@ public RepositorySettings RepositorySettings(Uri repositoryUri, string targetBra
                 RepositoryUri = repositoryUri,
                 RepositoryName = repoName,
                 RepositoryOwner = repoOwner,
+                RemoteInfo = ( targetBranch != null ? new RemoteInfo() { BranchName = $"origin/{targetBranch}" } : null )
```
MarcBruins: no need for the `(` and the `)`

Chama-StephanTuinder: removed

NuKeeper/Engine/RepositoryUpdater.cs
MarcBruins: We should only do a checkout if our currentBranch != targetBranch right? So if i don't use the targetBranch option this will checkout the branch that is already checked out, so basically doing it twice?

Chama-StephanTuinder: I will modify the clone to make add the branch info. That will fix that

NuKeeper.Abstractions/Git/IGitDriver.cs
```diff
@@ -9,6 +9,8 @@ public interface IGitDriver
 
         void Clone(Uri pullEndpoint);
 
+        void Clone(Uri pullEndpoint, string branchName);
```
MarcBruins: I think a clone function should mimic `git clone` as closely as possible. Therefore adding a extra parameter `branchName` doesn't make sense to me.

NuKeeper/Engine/RepositoryUpdater.cs
```diff
@@ -106,7 +106,7 @@ public class RepositoryUpdater : IRepositoryUpdater
 
         private static void GitInit(IGitDriver git, RepositoryData repository)
         {
-            git.Clone(repository.Pull.Uri);
+            git.Clone(repository.Pull.Uri, repository.DefaultBranch);
```
MarcBruins: Can't we just check here if `targetBranch is set then do a gitclone? Think you can just check if `repository.DefaultBranch !=  git.GetCurrentHead()`

Chama-StephanTuinder: repository.DefaultBranch != git.GetCurrentHead() only works áfter the clone
So then you are back to the extra checkout.
LibGit2Sharp supports the clone-with-branch out of the box

NuKeeper.Abstractions/Git/IGitDriver.cs
Chama-StephanTuinder: Git clone supports it:

something like:
git clone --single-branch --branch [branchname] [Git-Uri]

NuKeeper/Engine/RepositoryUpdater.cs
MarcBruins: Awesome i didn't know that

MarcBruins:
I'll merge this end of today if there are no comments from the other maintainers.

PR #756, opened 2019-04-25
Update gitea.md

MarcBruins:
fix headings


MarcBruins:
Seems like we have a flaky test `ConfigKeysAreCaseInsensitive`

PR #785, opened 2019-05-15
Updates Dockerfiles to NuKeeper release 0.21.0

skolima:
Current dockerfiles are 8 releases behind


MarcBruins:
LGTM

MarcBruins:
Like you said in gitter, we should automate this. Shouldn't be that hard.

We can add a docker task to the release pipeline that does a release build, then have a dockerfile that looks something like:

`
use dotnet sdk 2.1
set workdir to app
copy release binaries to container
entrypoint [dotnet, nukeeper.dll]
`

skolima:
I like the current approach of using the exact published version, though (by installing it), instead of  a second build.

Issue #799, opened 2019-05-25
Git commandline as Lib2git alternative


sharpSteff:
## 🚀 Feature Requests
Currently I'm facing some issues with Lib2Git(Sharp). It seems the git implementation differs from the native git implementation. Which results into bad (or not possible) repository interactions, e.g. when a (reverse) proxy is active or ssh-protocoll should be used.

I think it would be a great feature to give the user to possibility to enter the path to the git commandline as parameter in case the lib2git library is not enough

### Contextualize the feature
Second implementation of `IGitDriver `beside `LibGit2SharpDriver`

additional optional parameter `--gitcmdPath`, which is null per default

### Describe the feature
Implementation of a `GitCmdDriver : IGitDriver` which can be used as alternative to lib2sharpdriver.

Nukeeper should not ship git cmd, just give the option to use the installed git  

### Platforms affected (mark all that apply)
- [X] :hammer_and_wrench: NuKeeper CLI
- [X] :sparkles: GitHub
- [X] :robot: AzureDevops
- [X] :checkered_flag: Bitbucket
- [X] :earth_americas: Gitlab
- [X] :tv: Gitea
- [] :whale: Docker

sharpSteff:
started implementation on my fork
3b372111c875a9f4fdb01a508189ede03f162ed7

sharpSteff:
its crazy the proxy issue immediately disappeared when using git cli

sharpSteff:
I stumbled over #659. Any information how this is ongoing?

MaxMommersteeg:
I think this has been merged in. Can we close this issue?

shep1987:
Thanks for the work!

PR #800, opened 2019-05-25
Feature: Optional Git cmd support

sharpSteff:
Feature

No git cmd support

optional git cmd instead of Lib2GitSharp 

should not, since its an additional feature

Maybe provide a git cmd resource for unit testing.

All unit tests which does git interactions should also use the GitCmdDriver as IGitDriver

#799 

- [ ] All projects build


MarcBruins:
Cool! @skolima worked on this before and i think there where some issues he encountered

sharpSteff:
I tested it on gitea and GitHub. I think the biggest benifit of this Feature is, that you can rely on your already working.gitconfig in case of special stuff like proxy.

Do you have any Idea how you run unittests for this?

skolima:
#659 was extremely unreliable and I gave up - the unpredictability of this approach was not worth the gain (even though I have huge repositories where doing a shallow clone would be a huge speedup). There's little in terms of test coverage, though, so for the spike I was relying on running this by hand for various repositories. In order to be able to merge this, you'd need to write the tests as well (current NuKeeper libgit integration is lacking them - not ideal).

sharpSteff:
@skolima thanks for the Feedback. I see this Feature as a fallback solution in case lib2git does not the job. any ideas how to get git cmd running on build servers?

skolima:
What I've tried doing before was checking if %PATH% actually contains git already - this would be the most common situation, I think. If not, then the build definition would need to be changed to install it.

MarcBruins:
Two things i noticed is that the GitDiscoveryDriver isn't updated and there are no docs update with this PR.

sharpSteff:
Will fix that tomorrow

skolima:
In `StartGitProzess` (typo!) you've got your own `Process` handling. This should be using `ExternalProcess` as other places that launch executables from NuKeeper do. 2 reasons for this: a ) consistency b) if there's *any* bugs in `ExternalProcess`, those are rather critical to how NuKeeper works, so, we need to fix them. And if this works, then definitely the logging support should be used here as well.

skolima:
`ExternalProcess.Run` is `async`, you'd need (AFAIK) to `await` it, calling `.Result` directly leads to all kinds of problems. Which means the `async` keyword will need to propagate a bit up the call stack.

sharpSteff:
@skolima calling .Result is indeed a blocking call, but in this case there should be no problems.

However a better approach as you mentioned, to include `async`

sharpSteff:
@skolima managed to get rid of the `.Result`.

Also added some unit tests.
I still have to check whether everything runs also on Linux.
And in addition I have to finish the `GitDiscoveryDriver `

MarcBruins:
There are still merge conflicts can you check them? Are you also confident that this works now?

sharpSteff:
@MarcBruins 
there are still 2 open points:

-  update docu
-  The IGitDiscoveryDriver is currently as a singleton implemented. Any ideas to inject dynamically the GitCmdDiscoveryDriver, when the command line is applied.

sharpSteff:
updated docu

sharpSteff:
@MarcBruins I'm confident everything works. We're already running our nukeeper on windows with git cli (repo/update) commands. I tested everything as well on Ubuntu. Also I added unit tests to be sure.

Chama-StephanTuinder:
Would be nice to see a full-implementation of the GitCmdDriver.

StephanTuinder:
Would be nice to see a full-implementation of the GitCmdDriver.

sharpSteff:
@StephanTuinder 
What additional Features are needed ?

StephanTuinder:
Weird, in my mind there where a lot of NotImplemented routines. Probably my branch was behind.

sharpSteff:
Never mind!

Great that you find that issue in GitCmdDiscoveryDriver. I'll take a look at the tests

StephanTuinder:
I fixed it in my PR, so check there

PR #816, opened 2019-05-31
Remove side effects from setting readers

zubivan:

That's sort of a bug fix. Every time `GitHubSettingsReader` `CanRead` was called it's internal state was altered.

Internal state of presumably stateless class is altered.

It's not altered anymore. :)

It shouldn't introduce any.

Should work as it used to.

- [x] All projects build

### DIFF
--- a/NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
+++ b/NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
@@ -13,7 +13,6 @@ namespace NuKeeper.AzureDevOps
         private const string UrlPattern = "https://dev.azure.com/{org}/{project}/_git/{repo}/";
 
         private readonly IGitDiscoveryDriver _gitDriver;
-        private bool _isLocalGitRepo;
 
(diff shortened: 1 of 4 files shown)


NuKeeper.AzureDevOps/AzureDevOpsSettingsReader.cs
```diff
@@ -13,44 +13,35 @@ public class AzureDevOpsSettingsReader : BaseSettingsReader
...
             _gitDriver = gitDriver;
         }
 
         public override bool CanRead(Uri repositoryUri)
         {
-            if (repositoryUri == null)
-            {
-                return false;
-            }
+            if (repositoryUri == null) return false;
```
skolima: This goes against the style used in NuKeeper. We might need a style file to enforce this.

zubivan: Oh yeah, also noticed it. Accidentally clicked `Code cleanup` when using Rider. Will fix.

zubivan: BTW - it's fixed.

PR #825, opened 2019-06-23
Added support for mutliple item groups in Directory.Build.props/targets

david-driscoll:
Bug fix

Currently only the first `<ItemGroup>` is searched in a `Directory.Build.props` or `Directory.Build.targets` file.

Read and update all the `<PackageReference>`s in `Directory.Build.props` or `Directory.Build.targets`.

Not that I'm aware of.

Added 2 tests to the integration suite that demonstrate the problem.  Verified by creating the failing test first.

- [X] All projects build
Unsure if this needs updating in documentation.


david-driscoll:
Also added support for https://github.com/Microsoft/MSBuildSdks/tree/master/src/CentralPackageVersions  which basically just uses `Packages.props` in place of `Directory.Build.targets`.

Let me know if that should break out to another class.

MarcBruins:
Think this looks fine, very nice addition. Thanks you sir!

PR #831, opened 2019-06-27
Fix for Casing problems Nugets

StephanTuinder:
**Bug Fix:**
When someone adds a nuget with ```dotnet add``` and gets the casing wrong, it will be added with the wrong casing to the csproj file and Nukeeper won't recognize it for updates.

Example: 
```dotnet add package newtonsoft.json -v 8.0.1``` 
will add it as 
```<PackageReference Include="newtonsoft.json" Version="8.0.1" />```
to the .csproj file.
Visual Studio has no problem with that but NuKeeper tries to match it to "Newtonsoft.Json" and does not recognize it.

NuKeeper does not update nugets where the casing is wrong in the .csproj file

NuKeeper will update nugets where the casing is wrong in the .csproj file because the checks have been made case-insensitive.

No

See Description

No Issue (yet)

- [X] All projects build

### DIFF
--- a/NuKeeper.Abstractions/RepositoryInspection/PackageUpdateSet.cs
+++ b/NuKeeper.Abstractions/RepositoryInspection/PackageUpdateSet.cs
@@ -59,12 +59,12 @@ namespace NuKeeper.Abstractions.RepositoryInspection
 
         private void CheckIdConsistency()
         {
-            if (CurrentPackages.Any(p => p.Id != SelectedId))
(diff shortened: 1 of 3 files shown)


zubivan:
Quite a stupid behaviour of `dotnet add` to be fair. Feels like it should respect the casing defined by package author when saving it in the project file. Maybe create an issue for that?

Chama-StephanTuinder:
The solution proposed should make it foolproof  (until they invent a bigger fool  😄 )

StephanTuinder:
The solution proposed should make it foolproof (until they invent a bigger fool 😄 )

PR #833, opened 2019-06-27
Support for transistive dependencies.

MaxMommersteeg:
Can someone put the `Work In Progress` label on here?

Feature.

Currently, NuKeeper performs a `dotnet restore` per project it is going to create a package update for.
It also performs the package update using `dotnet add package` which by default does a restore preview. Doing both of these restore's when in the middle of updating transistive packages will result in a failing restore and thus failing NuKeeper execution.

NuKeeper should now support transistive dependencies.

We by default no longer perform a `dotnet restore` before updating a `ProjectFileOldStyle` or `ProjectFile` project. However this behavior can be toggled with the new UserSetting: `RestoreBeforePackageUpdate`.

We now also set the `--no-restore` flag when performing the package update with `dotnet add package`, this prevents an additional restore.

After doing the package updates we are still backtracking what `ProjectFileOldStyle` or `ProjectFile` projects received package updates and we will then perform the `dotnet restore` per project at the end of the run.

Yes, but it shouldn't 'hurt'. We changed the way we are performing dotnet restore's.

I still need to write tests for the new and changed functionality.

We discussed this feature at the dutch Hackathon (2019/06/25).

- [x] All projects build

### DIFF
--- a/NuKeeper.Abstractions/Configuration/ISettingsContainer.cs
+++ b/NuKeeper.Abstractions/Configuration/ISettingsContainer.cs
@@ -1,0 +1,17 @@
+using NuKeeper.Abstractions.Inspections.Files;
+
+namespace NuKeeper.Abstractions.Configuration
+{
+    public interface ISettingsContainer
+    {
+        SourceControlServerSettings SourceControlServerSettings { get; set; }
+
+        UserSettings UserSettings { get; set; }
+
+        FilterSettings PackageFilters { get; set; }
+
+        BranchSettings BranchSettings { get; set; }
+
+        IFolder WorkingFolder { get; set; }
+    }
+}
--- a/NuKeeper.Update.Tests/UpdateRunnerTests.cs
+++ b/NuKeeper.Update.Tests/UpdateRunnerTests.cs
@@ -52,7 +52,7 @@ namespace NuKeeper.Update.Tests
                 _updateDirectoryBuildTargetsCommand);
         }
 
-        [TestCase(true, Description = "Should also perform dotnet restore's before each package update.")]
+        [TestCase(true, Description = "Should perform dotnet restore's before each package update.")]
         [TestCase(false, Description = "Should not perform dotnet restore's before each package update.")]
         public async Task CorrectCommandsAreExecutedForProjectFileUpdate(bool restoreBeforePackageUpdate)
         {
@@ -68,7 +68,7 @@ namespace NuKeeper.Update.Tests
             await AssertCorrectProjectFileCommandsAreExecuted(packageUpdateSet, sources, restoreBeforePackageUpdate);
         }
 
-        [TestCase(true, Description = "Should also perform dotnet restore's before each package update.")]
+        [TestCase(true, Description = "Should perform dotnet restore's before each package update.")]
         [TestCase(false, Description = "Should not perform dotnet restore's before each package update.")]
         public async Task CorrectCommandsAreExecutedForProjectFileOldStyleUpdate(bool restoreBeforePackageUpdate)
         {
--- a/NuKeeper.Update.Tests/UpdateRunnerTests.cs
+++ b/NuKeeper.Update.Tests/UpdateRunnerTests.cs
@@ -1,0 +1,243 @@
+using System;
+using System.Collections.Generic;
+using NuGet.Configuration;
+using NuGet.Packaging.Core;
+using NuGet.Versioning;
+using NuKeeper.Abstractions.Configuration;
+using NuKeeper.Abstractions.NuGetApi;
+using NuKeeper.Abstractions.RepositoryInspection;
+using NSubstitute;
+using NuKeeper.Abstractions.Logging;
+using NuKeeper.Update.Process;
+using NUnit.Framework;
+using NuKeeper.Abstractions.NuGet;
+using System.Threading.Tasks;
+using System.IO;
+
+namespace NuKeeper.Update.Tests
+{
+    [TestFixture]
+    public class UpdateRunnerTests
+    {
+        private INuKeeperLogger _nuKeeperLogger;
+        private IFileRestoreCommand _fileRestoreCommand;
+        private INuGetUpdatePackageCommand _nuGetUpdatePackageCommand;
+        private IDotNetUpdatePackageCommand _dotNetUpdatePackageCommand;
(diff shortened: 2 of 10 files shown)


MaxMommersteeg:
I added unittests to validate the amount and type of commands that get executed by the `UpdateRunner` class. Two tests fail since we still need to inject and use that `RestoreBeforePackageUpdate` boolean.

MaxMommersteeg:
Still stuck at injection settings into the UpdateRunner. Any suggestions would be appreciated.

NuKeeper.Abstractions/Configuration/ISettingsContainer.cs
```diff
@@ -0,0 +1,17 @@
+using NuKeeper.Abstractions.Inspections.Files;
+
+namespace NuKeeper.Abstractions.Configuration
+{
+    public interface ISettingsContainer
```
shep1987: Shouldn't need an interface for a settings object should just be passed into things

NuKeeper.Update.Tests/UpdateRunnerTests.cs
```diff
@@ -0,0 +1,218 @@
...
+        private IUpdateNuspecCommand _updateNuspecCommand;
+        private IUpdateDirectoryBuildTargetsCommand _updateDirectoryBuildTargetsCommand;
+        private IUpdateRunner _sut;
+
+        [OneTimeSetUp]
+        public void SetUp()
+        {
+            _nuKeeperLogger = Substitute.For<INuKeeperLogger>();
+
+            _settingsContainer = Substitute.For<ISettingsContainer>();
```
shep1987: Just create the object here with defaults, you can then change settings in each test class

NuKeeper.Update/UpdateRunner.cs
```diff
@@ -46,14 +53,20 @@ public async Task Update(PackageUpdateSet updateSet, NuGetSources sources)
 
             foreach (var current in sortedUpdates)
             {
-                var updateCommands = GetUpdateCommands(current.Path.PackageReferenceType);
+                var updateCommands = GetUpdateCommands(current.Path.PackageReferenceType, _settingsContainer.UserSettings.RestoreBeforePackageUpdate);
```
shep1987: Add the RestoreBeforePackageUpdate param to the method signature

NuKeeper.Update.Tests/UpdateRunnerTests.cs
```diff
@@ -0,0 +1,218 @@
...
+        public async Task CorrectCommandsAreExecutedForProjectFileUpdate(bool restoreBeforePackageUpdate)
+        {
+            // Arrange
+            _settingsContainer.UserSettings.RestoreBeforePackageUpdate = restoreBeforePackageUpdate;
+            var packageReferenceType = PackageReferenceType.ProjectFile;
+            var packageUpdateSet = UpdateFooFromOneVersion(packageReferenceType);
+            var sources = NuGetSources.GlobalFeed;
+
+            // Act
+            await _sut.Update(packageUpdateSet, sources);
```
shep1987: this would then take in restoreBeforePackageUpdate rather than the _settings container

NuKeeper.Update/Process/DotNetUpdatePackageCommand.cs
```diff
@@ -22,18 +22,14 @@ public DotNetUpdatePackageCommand(IExternalProcess externalProcess)
...
-            await _externalProcess.Run(projectPath, "dotnet", restoreCommand, true);
 
             if (currentPackage.Path.PackageReferenceType == PackageReferenceType.ProjectFileOldStyle)
             {
                 var removeCommand = $"remove {projectFileName} package {currentPackage.Id}";
                 await _externalProcess.Run(projectPath, "dotnet", removeCommand, true);
             }
 
-            var addCommand = $"add {projectFileName} package {currentPackage.Id} -v {newVersion} -s {sourceUrl}";
+            var addCommand = $"add {projectFileName} package {currentPackage.Id} -v {newVersion} -s {sourceUrl} --no-restore";
```
shep1987: You could create 2 implementations of the interface then create a factory to return you one with --no-restore on or not, this would be a safer change to make as the implementation would then be identical

NuKeeper.Update/UpdateRunner.cs
```diff
@@ -64,37 +77,56 @@ private IReadOnlyCollection<PackageInProject> Sort(IReadOnlyCollection<PackageIn
         }
 
         private IReadOnlyCollection<IPackageCommand> GetUpdateCommands(
```
shep1987: If you go for factory approach you would do something like
var _dotNetUpdatePackageCommand = _dotNetUpdatePackageCommandFactory.GetDotNetUpdatePackageCommand(true/false)

then at the end add the DotNetRestoreCommand  if needed

SuddenGunter:
@MaxMommersteeg I stumbled upon same issue as you.
At the moment I've just patched it locally in commenting lines 27 and 28 here https://github.com/NuKeeperDotNet/NuKeeper/blob/bc357aaeacbaaad40c422091fcd7d26ae402db7b/NuKeeper.Update/Process/DotNetUpdatePackageCommand.cs#L27 

but it's just a temporary fix and I suggest we fix this for whole community.
@MaxMommersteeg @shep1987 is there are a way this PR can be fixed and finished?
 or this is up for grabs?
we really need this feature and dirty patch not a good solution in a long run. 
@MaxMommersteeg if you are no longer interested in finishing it I can do it

MaxMommersteeg:
Thanks for your comment @SuddenGunter.  Sadly I didn't have the time to implement the suggestions given by @shep1987 . Getting this code in production would help out many others facing this issue.

Helping me finish this PR would definitely help. Be my guest :)

SuddenGunter:
@MaxMommersteeg ok, thanks)
@shep1987 what was the main reason to restore package on each package update, and not after all updates?

jonathann92:
Hi guys, is this PR still being considered to be merged in? This PR would benefit my automation of keeping nuget packages up to date :)

SuddenGunter:
@jonathann92 I fixed it for my particular project with this commit: https://github.com/dashdevs/NuKeeper/commit/abf6b3db8f512b476244cab205e47e446cb62518

It's a dirty hack, but it works

PR #836, opened 2019-06-28
Improver Exception handling Octokit (GitHub). Exceptions will show more info than the current "Validation Error"

StephanTuinder:
feature (kid of)

When Octokit throws an error, the only thing logged is "Validation Error" which is not much help.

The Body of the http response (if there) is parsed to see if the real error can be determined.
If errors can be found in the body, the first will be in the message of the Exception thrown.
The original Exception is passed to the thrown Exception als innerException.
The thrown exception will be a NuKeeper exception.

no

As an example you could try to run nukeeper on a repository that you only have read access to.

- [X] All projects build


StephanTuinder:
I know it leads to duplication, but I don't see how I could create a global exeption handler for this class alone. I will do some research.

StephanTuinder:
Added global error handler for class

Issue #843, opened 2019-07-04
BitbucketLocal doesn't work


evolvedlight:
## 🐛 Bug Report

It seems that this doesn't work anymore thanks to a new validation.
It just gives the error ""

### Expected behavior
As delete after merge isn't supported, it doesn't validate or try to do it

### Reproduction steps

λ NuKeeper.exe repo --platform bitbucketlocal http://bitbucket.ourcompany.com/scm/project/gitrepo.git token
Collaboration platform specified as 'BitbucketLocal'
Deletion of source branch after merge is currently only available for Azure DevOps, Gitlab and Bitbucket.

### Configuration

**Version:** 0.24.0

**Platform if applicable:** 
- [x ] :hammer_and_wrench: NuKeeper CLI
- [x ] :checkered_flag: Bitbucket

msallin:
As far as I see, there are two problems here.
1) The default value for `DeleteBranchAfterMerge` is `true`, also if the target platform doesn't support this.
2) The validation is broken, therefore even if `DeleteBranchAfterMerge` is set to `false`, none of the not supported platforms are going to work.

I created a PR to fix it. Didn't had time to create Unit Tests but the method `PopulateDeleteBranchAfterMerge` should be tested well, as there are a lot of possible states which can occurred. Feel free to contribute: https://github.com/NuKeeperDotNet/NuKeeper/pull/844

zubivan:
@msallin @evolvedlight should be resolved now.

msallin:
Tested. It works.

zubivan:
@msallin Thank you for taking a look!

Issue #798, opened 2019-05-22
Platform argument not followed unless repo URL matches satisfies the string pattern


ElanHasson:
## 🔙 Regression

`Platform` config value is ignored unless the repository URL matches the string pattern in the platform `ISettingsReader` implementation.

### Old (and correct) behavior
Doesn't seem to have worked.

### Current behavior
Platform is ignored unless Repository URL matches the string pattern.

### Reproduction steps
`nukeeper repo https://myserver/projects/org/reponame token` fails because the URL doesn't match any of the platform `ISettingsReader` implementation.
`nukeeper repo --platform Bitbucketlocal https://myserver/projects/org/reponame token`  fails because there is no logic to short circuit the string pattern matching if platform is specified.

### Configuration

**Version:** 0.21.2

**Platform if applicable:** 
- [X] :hammer_and_wrench: NuKeeper CLI
- [X] :sparkles: GitHub
- [X] :robot: AzureDevops
- [X] :checkered_flag: Bitbucket
- [X] :earth_americas: Gitlab
- [X] :tv: Gitea
- [X] :whale: Docker

ElanHasson:
See #590 also.

jwmoffat:
I can confirm this as well.  

I've been trying to get it to work on version 0.21.3 with gitlab.  

Without any flag, I get the error "Unable to work out which platform to use [URL] could not be matched"

At first I thought I was setting the platform flag incorrectly, but if you give it an invalid option it prints out "Invalid value specified for platform. Allowed values are: GitHub, AzureDevOps, Bitbucket, BitbucketLocal, GitLab, Gitea"

However, even with a valid platform flag, I still get the same original error.

Changing my URL (via a hosts entry in Windows) to include "gitlab" in the URL then allows it to move forward.

msallin:
There is currently an issue with our local Bitbucket installation, which prevents me from testing and I'm also not sure if I catched the root of the issue. Please review and test.

https://github.com/NuKeeperDotNet/NuKeeper/pull/821

zubivan:
Also observed this behaviour, great that you're fixing it!

StephanTuinder:
My 2 cents: It seems to be ignored in the "repo" mode

msallin:
@ElanHasson close the issue?

ElanHasson:
I haven't had a chance to revisit this. I'll close for now and reopen if I see it again

PR #844, opened 2019-07-08
#843 Correct validation of DeleteBranchAfterMerge

msallin:
Bug

https://github.com/NuKeeperDotNet/NuKeeper/issues/843

https://github.com/NuKeeperDotNet/NuKeeper/issues/843

No.

Unit Tests missing.

https://github.com/NuKeeperDotNet/NuKeeper/issues/843

- [x] All projects build

### DIFF
--- a/NuKeeper/Commands/CollaborationPlatformCommand.cs
+++ b/NuKeeper/Commands/CollaborationPlatformCommand.cs
@@ -50,12 +50,17 @@ namespace NuKeeper.Commands
             Description = "Deletes branch created by NuKeeper after merge. Defaults to true.")]
         public bool? DeleteBranchAfterMerge { get; set; }
 
+        private HashSet<Platform> _platformsSupportingDeleteBranchAfterMerge = new HashSet<Platform>();
+
         protected CollaborationPlatformCommand(ICollaborationEngine engine, IConfigureLogger logger,
             IFileSettingsCache fileSettingsCache, ICollaborationFactory collaborationFactory) :
             base(logger, fileSettingsCache)
         {
             _engine = engine;
             CollaborationFactory = collaborationFactory;
+            _platformsSupportingDeleteBranchAfterMerge.Add(Abstractions.CollaborationPlatform.Platform.AzureDevOps);
+            _platformsSupportingDeleteBranchAfterMerge.Add(Abstractions.CollaborationPlatform.Platform.Bitbucket);
+            _platformsSupportingDeleteBranchAfterMerge.Add(Abstractions.CollaborationPlatform.Platform.GitHub);
         }
 
         protected override async Task<ValidationResult> PopulateSettings(SettingsContainer settings)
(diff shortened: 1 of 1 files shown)


NuKeeper/Commands/CollaborationPlatformCommand.cs
```diff
@@ -132,21 +132,34 @@ protected override async Task<int> Run(SettingsContainer settings)
...
 
-            if (!Platform.HasValue)
+
+            bool defaultValue;
+
+            // The default value is true, if it is supported for the corresponding platform.
+            if (Platform.HasValue &&
+                Platform != Abstractions.CollaborationPlatform.Platform.AzureDevOps
+                && Platform != Abstractions.CollaborationPlatform.Platform.GitLab
+                && Platform != Abstractions.CollaborationPlatform.Platform.Bitbucket)
```
evolvedlight: Is this the right way around? Bitbucket (I believe) supports delete after merge, bitbucket local doesn't support it. Might be good to move this list to be shared with lines 158-158 too? Thanks :)

msallin: Bitbucket/BitbucketLocal:
This is true, Bitbucket does support it and BitbucketLocal doesn't. 

Right way around:
Yes, it is the right way around as this condition is true when there is a platform specified and this platform is NOT one of those tthree. Thus, the default value is going to be false in case of an unsupported platform.

Shared List:
Yes, I'll do it.

msallin: Btw. a general question&remark which arised while working on #798 & #843
It is currently hard to work with platform related things as it is not possible to see what platform is used, unless it was specified explicitly. This is also the case there. It is still possible that the wrong default value is used, if a platform was chosen from the URL.
I don't have a good overview over the code base but from what I saw, it would make sense to not tie the selection of the platform to the Settings reader but rather do it before the settings reader is chosen.
Concrete, that means, evaluating the platform first (by Url, passed argument or file) and set the correct enum value. Then it doesn't matter anymore from where this value originated.

NuKeeper/Commands/CollaborationPlatformCommand.cs
```diff
@@ -50,12 +50,17 @@ internal abstract class CollaborationPlatformCommand : CommandBase
...
+
         protected CollaborationPlatformCommand(ICollaborationEngine engine, IConfigureLogger logger,
             IFileSettingsCache fileSettingsCache, ICollaborationFactory collaborationFactory) :
             base(logger, fileSettingsCache)
         {
             _engine = engine;
             CollaborationFactory = collaborationFactory;
+            _platformsSupportingDeleteBranchAfterMerge.Add(Abstractions.CollaborationPlatform.Platform.AzureDevOps);
+            _platformsSupportingDeleteBranchAfterMerge.Add(Abstractions.CollaborationPlatform.Platform.Bitbucket);
+            _platformsSupportingDeleteBranchAfterMerge.Add(Abstractions.CollaborationPlatform.Platform.GitHub);
```
Maxwe11: You meant GilLab ?

Maxwe11:
Any chance to get this merged?

NuKeeper/Commands/CollaborationPlatformCommand.cs
msallin: Yes. Thank you! I'll correct it.

msallin:
Hi guys.
Without this fix BitbucketLocal does not work. It would be great to get it merged soon. Is there anything left, which I've to do?

NuKeeper/Commands/CollaborationPlatformCommand.cs
```diff
@@ -132,21 +137,29 @@ protected override async Task<int> Run(SettingsContainer settings)
         {
             var fileSettings = FileSettingsCache.GetSettings();
 
-            if (!Platform.HasValue)
+
```
zubivan: I think that extra line is not really necessary.

PR #845, opened 2019-07-08
When there are errors during the Integration tests, the Nukeeper and Nuget log is put into the test output

StephanTuinder:
In the Integration tests, some tests are sometimes failing without a clear reason.
With this change, whenever a test fails, it wil put the output of the logging into the output of the test.

No logging available

Logging available

No

Fail a test

- [X] All projects build

### DIFF
--- a/NuKeeper.Integration.Tests/BaseTest.cs
+++ b/NuKeeper.Integration.Tests/BaseTest.cs
@@ -1,0 +1,32 @@
+using NuGet.Common;
+using NuKeeper.Abstractions.Logging;
+using NuKeeper.Integration.Tests.LogHelpers;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+using System;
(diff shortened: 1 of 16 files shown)


MarcBruins:
This is adding complexity and more logic to the tests which i think in general is bad pratice. Shouldn't we focus on the test that fail without clear reasons rather then to add logging to tests?

StephanTuinder:
The problem with the methods under test is that they mostly return a single value:  true/false etc.
To make it so that the test results are more clear, we have to change the methods under test.
The idea of adding the logging is that by implementing a special logger the tests are not impacted, but it will still be more clear what the flow is that results in failing the test.

AnthonySteele:
Maybe a naming change? `public class BaseTest` to `public abstract class TestWithFailureLogging` ?
I general purpose "BaseTest" with all the unrelated things in it is a pattern to avoid.

PR #868, opened 2019-08-28
Initialize NuGet's Credential Service

ransagy:

Somewhere between feature and bug fix.

When performing NuGet operations against private/authenticated feeds, Current behavior assumes that auth is done solely via basic http authentication. This is not always true in some cases (Azure DevOps Artifacts in my case) and NuGet has a plugin framework for credentials management.

We initialize NuGet's credentials service before performing operations, allowing the usage of NuGet's own providers for that. In my case, Microsoft's Azure Artifacts credentials provider.

Possibly, If a user has credentials providers installed that will override the old auth system.

The auth tests i saw seem to cover these cases, But we might consider adding more tests that specifically use the credential providers and/or Azure Artifacts authentication. It might be too much of an overhead though.

https://docs.microsoft.com/en-us/nuget/reference/extensibility/nuget-exe-credential-providers
https://github.com/microsoft/artifacts-credprovider

#542 

- [x] All projects build
- [ ] Relevant documentation was updated (Not sure what constitutes relevant in this case - I'll be happy to do so with some pointers)

### DIFF
--- a/NuKeeper/Local/LocalEngine.cs
+++ b/NuKeeper/Local/LocalEngine.cs
@@ -3,6 +3,7 @@ using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using NuGet.Common;
 using NuGet.Credentials;
 using NuKeeper.Abstractions.Configuration;
 using NuKeeper.Abstractions.Inspections.Files;
@@ -25,6 +26,7 @@ namespace NuKeeper.Local
         private readonly ILocalUpdater _updater;
         private readonly IReporter _reporter;
         private readonly INuKeeperLogger _logger;
+        private readonly ILogger _nugetLogger;
 
         public LocalEngine(
             INuGetSourcesReader nuGetSourcesReader,
@@ -32,7 +34,8 @@ namespace NuKeeper.Local
             IPackageUpdateSetSort sorter,
             ILocalUpdater updater,
             IReporter reporter,
-            INuKeeperLogger logger)
+            INuKeeperLogger logger,
+            ILogger nugetLogger)
         {
             _nuGetSourcesReader = nuGetSourcesReader;
             _updateFinder = updateFinder;
@@ -40,11 +43,12 @@ namespace NuKeeper.Local
             _updater = updater;
             _reporter = reporter;
             _logger = logger;
+            _nugetLogger = nugetLogger;
         }
 
         public async Task Run(SettingsContainer settings, bool write)
         {
-            DefaultCredentialServiceUtility.SetupDefaultCredentialService(new NuGet.Common.NullLogger(), true);
+            DefaultCredentialServiceUtility.SetupDefaultCredentialService(_nugetLogger, true);
 
             var folder = TargetFolder(settings.UserSettings);
 
--- a/NuKeeper/Local/LocalEngine.cs
+++ b/NuKeeper/Local/LocalEngine.cs
@@ -3,6 +3,7 @@ using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using NuGet.Credentials;
 using NuKeeper.Abstractions.Configuration;
 using NuKeeper.Abstractions.Inspections.Files;
 using NuKeeper.Abstractions.Logging;
@@ -43,6 +44,8 @@ namespace NuKeeper.Local
 
         public async Task Run(SettingsContainer settings, bool write)
         {
+            DefaultCredentialServiceUtility.SetupDefaultCredentialService(new NuGet.Common.NullLogger(), true);
+
             var folder = TargetFolder(settings.UserSettings);
 
             var sources = _nuGetSourcesReader.Read(folder, settings.UserSettings.NuGetSources);
--- a/NuKeeper.Tests/Local/LocalEngineTests.cs
+++ b/NuKeeper.Tests/Local/LocalEngineTests.cs
@@ -90,10 +90,12 @@ namespace NuKeeper.Tests.Local
 
             var logger = Substitute.For<INuKeeperLogger>();
 
-            var reporter = Substitute.For<IReporter>();
+            var nugetLogger = Substitute.For<NuGet.Common.ILogger>();
 
+            var reporter = Substitute.For<IReporter>();
+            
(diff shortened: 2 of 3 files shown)


NuKeeper/Local/LocalEngine.cs
```diff
@@ -43,6 +44,8 @@ public class LocalEngine : ILocalEngine
 
         public async Task Run(SettingsContainer settings, bool write)
         {
+            DefaultCredentialServiceUtility.SetupDefaultCredentialService(new NuGet.Common.NullLogger(), true);
```
ransagy: I'm definitely open to suggestions here; Wasn't sure where the best place is to put this.
It's also possible we want to use a different logger and/or to do something to integrate NuKeeper's logger with the basic NuGet one?

skolima: If you'd add a dependency of `NuGet.Common.ILogger` to the constructor, you'll get one injected that will forward the events to the NuKeeper logging.

Other than that, I think this looks good (other than the fact it's awkward static setup, but that's how this code works in NuGet itself and not much we can do about it).

ransagy: I'll fix that ASAP then, Thanks.

MartinDemberger: I'm not sure about this but should this Setup-method also be called in CollaborationEngine?

ransagy: @skolima Thanks to @Greybird this is now fixed.
As per @MartinDemberger 's comment, I'd need another pair of eyes on that.

ransagy:
Tests are broken due to a non-related issue to this PR; see my PR to fix that in #894 .

MarcBruins:
Is this ready to be reviewed again @ransagy?

ransagy:
Pending any additional comment in the thread about `LocalEngine.cs` but generally, yes indeed.

nullabletype:
Any movement on this? It's something I've just run into as well 🙃

ransagy:
Still waiting for approval/further review by the maintainers, as specified.
@MarcBruins can someone take a look? IIRC @skolima no longer reviews these as per some of comments on other open issues here.

bacpcgus:
Does anybody know when this might be released to a version?  Running into this on our Pipeline builds and it's a show-stopper.

scp-mb:
Also running up against this. Getting nukeeper to work via a devops pipeline either through the extension or a powershell script is headache inducing.

@AnthonySteele Any chance we can get an update on when to expect a new release?

josundt:
I am also awaiting this to be released. When can I expect a release that includes this?

ransagy:
I'm quite sure it was already included in the 0.28 and 0.29 NuGet releases; Just wasn't updated on the GitHub releases due to a bug, i think.
I'm not aware of the semantics needed for the pipeline scenario, Is there anything else required besides the tool being updated on NuGet?

AnthonySteele:
That's correct, it's on the last NuGet releases, but there was an issue with release notes.
I will look into updating the ADO Extensions with the next release.

NuKeeper/Local/LocalEngine.cs
johntyrrell: @ransagy I think I need you're fix in the CollaborationEngine. When I call runkeeper repo on a Azure devops nuget feed it can't find any packages. If my repo doesn't have a NuGet.config and I set credentials in my profile NuGet.config it works. But if my repository does have a NuGet.config without credentials it fails. 
To bypass this I call nukeeper with --source, but this too fails because of credentials.

I tried to update GitRepositoryEngine with the line 
DefaultCredentialServiceUtility.SetupDefaultCredentialService(_nugetLogger, true);
but when executing nukeeper I get the error:
Problem starting the plugin 'C:\Users\%username%\.nuget\plugins\netcore\CredentialProvider.Microsoft\CredentialProvider.Microsoft.dll'. Cannot start process because a file name has not been provided.

Can you help?

ransagy: The latter error sounds like the missing dotnet env path var: https://github.com/NuGet/Home/issues/7438

johntyrrell: @ransagy when I set the DOTNET_HOST_PATH like this:
[System.Environment]::SetEnvironmentVariable("DOTNET_HOST_PATH", "C:\Program Files\dotnet")

The error changes to this:
Problem starting the plugin 'C:\Users\%USERNAME%\.nuget\plugins\netcore\CredentialProvider.Microsoft\CredentialProvider.Microsoft.dll'. Access is denied.
System.ComponentModel.Win32Exception (5): Access is denied.
   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start()
   at NuGet.Protocol.Plugins.PluginProcess.Start()
   at NuGet.Protocol.Plugins.PluginFactory.CreatePluginAsync(String filePath, IEnumerable`1 arguments, IRequestHandlers requestHandlers, ConnectionOptions options, CancellationToken sessionCancellationToken)
   at NuGet.Protocol.Plugins.PluginFactory.GetOrCreateAsync(String filePath, IEnumerable`1 arguments, IRequestHandlers requestHandlers, ConnectionOptions options, CancellationToken sessionCancellationToken)
   at NuGet.Protocol.Plugins.PluginManager.TryCreatePluginAsync(PluginDiscoveryResult result, OperationClaim requestedOperationClaim, PluginRequestKey requestKey, String packageSourceRepository, JObject serviceIndex, CancellationToken cancellationToken)

Any ideas?

ransagy: I think that env var needs to be the full path to the actual dotnet exe - In my (local machine) case i just set it to 'dotnet' so it will use whatever is installed in the path.

About permissions - No idea, I haven't used the pipeline scenario at all.

johntyrrell: @ransagy , changing it do "dotnet" did the trick. After that I needed to also set NUGET_PLUGIN_HANDSHAKE_TIMEOUT_IN_SECONDS and NUGET_PLUGIN_REQUEST_TIMEOUT_IN_SECONDS to "30" and then I was able to read the Azure DevOps feed. Thnx!

PR #895, opened 2019-11-01
Added support for PackageDownload items

david-driscoll:
Feature!

Currently `PackageDownload` is not a supported target as supported element that can contain a version.  (This is seriously a thing)

Adds support for `PackageDownload`. The behavior is very similar to `PackageReference` except that the version must be defined as an exact version `[1.0.0]` instead of just `1.0.0`.

No

Testing with and without `PackageDownload`.

This is seriously a thing, today Nuke is using it ensure that packages are downloaded, but they are not implicitly referenced.  This is used in the Nuke context to ensure that the binaries exist.  I could not however find any specific documentation on this behavior, but it is a thing.

https://github.com/NuGet/NuGet.Client/blob/f5916a527c442f896f787c593252b37a345da835/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets#L845-L850

- [X] All projects build
  * Are there any docs to update around this?

### DIFF
--- a/NuKeeper.Inspection.Tests/RepositoryInspection/DirectoryBuildTargetsReaderTests.cs
+++ b/NuKeeper.Inspection.Tests/RepositoryInspection/DirectoryBuildTargetsReaderTests.cs
@@ -14,12 +14,19 @@ namespace NuKeeper.Inspection.Tests.RepositoryInspection
     [TestFixture]
     public class DirectoryBuildTargetsReaderTests
     {
-        const string PackagesFileWithSinglePackage =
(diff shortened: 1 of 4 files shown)


AnthonySteele:
We are looking into this PR
Apologies for the delay, we are picking up again and clearing stale issues, but keeping this one.

skolima:
NuGet implementation details: https://github.com/NuGet/Home/issues/7339

david-driscoll:
Closing in favor of #986

Issue #904, opened 2019-12-10
Executable isn't returning failure code when a failure occurs


sapeavy:
I'm using NuKeeper as a task within Azure DevOps. When the executable never gives a failure return code the task doesn't fail and I never know when my automated process needs attention.

Bouke:
Just to document some failure modes, here's a failure when restoring packages that returns a 0 exit code:
```
Updating 'Microsoft.CodeAnalysis.VersionCheckAnalyzer' from 2.9.4 to 2.9.8 in 1 projects
Nuget restore on d:\a\1\s\Source\NODEJSPROJECT packages.config
Updates failed NuKeeperException : Command C:\hostedtoolcache\windows\nukeeper\0.26.0\x64\NuGet.exe failed with exit code: 1

MSBuild auto-detection: using msbuild version '15.9.21.664' from 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\bin'.


Unable to locate project file for 'd:\a\1\s\Source\NODEJSPROJECT\packages.config'.

Attempted 3 updates and did 0
##[section]Finishing: NuKeeper
```

CrispyDrone:
For collaboration commands (`repo`, `global`, `organization`), this should now be fixed by https://github.com/NuKeeperDotNet/NuKeeper/pull/1058

PR #910, opened 2019-12-13
Return non-zero exit code for unhandled exceptions

Bouke:

feature

Currently when an exception is thrown besides `CommandParsingException`, the exit code of the tool will be 0 (no error). This means that in CI builds on a schedule, you won't get build failures if something is wrong (#904).

```
C:\Program Files\dotnet\dotnet.exe (process 13168) exited with code 0.
```

With this change, a non-zero exit code is returned, which would then fail the build and cause alarm bells to ring.

```
C:\Program Files\dotnet\dotnet.exe (process 13168) exited with code 2.
```

Maybe. So other tools relying on a zero exit code even when errors happen, would need a change. I think this is desirable.

Nuget.config:
```
@@ -0,0 +1,9 @@
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear/>
  </packageSources>
</configuration>
```

#904

- [ ] All projects build

### DIFF
--- a/NuKeeper/ExitCodes.cs
+++ b/NuKeeper/ExitCodes.cs
@@ -6,7 +6,7 @@ namespace NuKeeper
     enum ExitCodes
     {
         Success = 0,
-        UnknownError = 1 << 1,
-        InvalidArguments = 1 << 2,
+        UnknownError = 1 << 0,
+        InvalidArguments = 1 << 1,
     }
 }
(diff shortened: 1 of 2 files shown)


MarcBruins:
Mm the returning of a code other then `0` is definetly a good idea. I don't like the enum status codes and the additional cast that comes with it though. Can't we just return the integer value?

Bouke:
The reason for using enums is to be clear about which error codes can be returned, and a single location of defining those return codes. Using different error codes allows scripts to behave differently depending on the error code, log error code etc. The MVP here would be to return non-zero on errors, so dropping the enum is fine by me.

Issue #905, opened 2019-12-11
Error when running repo against private GitLab instance


paulowens:
## 🐛 Bug Report

Running repo command against a private instance of Gitlab returns the error, 'The provided uri was is not in the correct format.'

### Reproduction steps
Running the following command:
```$ nukeeper repo https://git.acmeaws.io/international/tools/nukeeper-test.git sadfdfddfd --api https://git.acmeaws.io/api/v4 --platform GitLab```

results in the following output:

```
Unhandled Exception: NuKeeper.Abstractions.NuKeeperException: The provided uri was is not in the correct format. Provided https://git.tmaws.io/intl-resale/tools/nukeeper-test.git and format should be https://gitlab.com/{username}/{projectname}.git
   at NuKeeper.Gitlab.GitlabSettingsReader.RepositorySettings(Uri repositoryUri, String targetBranch) in d:\a\r1\a\drop\NuKeeper.Gitlab\GitlabSettingsReader.cs:line 59
   at NuKeeper.Commands.RepositoryCommand.PopulateSettings(SettingsContainer settings) in d:\a\r1\a\drop\NuKeeper\Commands\RepositoryCommand.cs:line 57
   at NuKeeper.Commands.CommandBase.OnExecute() in d:\a\r1\a\drop\NuKeeper\Commands\CommandBase.cs:line 96
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.InvokeAsync(MethodInfo method, Object instance, Object[] arguments)
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.OnExecute(ConventionContext context, CancellationToken cancellationToken)
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.<>c__DisplayClass0_0.<<Apply>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.ExecuteAsync(String[] args, CancellationToken cancellationToken)
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.Execute(String[] args)
   at NuKeeper.Program.Main(String[] args) in d:\a\r1\a\drop\NuKeeper\Program.cs:line 33
```

### Configuration

**Version:** 0.26.0

**Platform if applicable:** 
- [ ] :hammer_and_wrench: NuKeeper CLI
[...]

PR #915, opened 2019-12-23
Add subgroup support to GitLab repositories

Calytic:

Bug fix

Specifying a repository uri with a single or multiple subgroup(s) throws an exception.

Specifying a repository uri with a single or multiple subgroup(s) does not throw an exception.

No.

Specify a repository with and without a subgroup or multiple subgroups.

#905 
https://nukeeper.com/commands/repository/
https://nukeeper.com/platform/gitlab/

- [ ] All projects build

### DIFF
--- a/NuKeeper.Gitlab/GitlabSettingsReader.cs
+++ b/NuKeeper.Gitlab/GitlabSettingsReader.cs
@@ -44,20 +44,20 @@ namespace NuKeeper.Gitlab
                     $"The provided uri was is not in the correct format. Provided null and format should be {UrlPattern}");
             }
 
-            // Assumption - url should look like https://gitlab.com/{username}/{projectname}.git";
+            // Assumption - url should look like https://gitlab.com/{username}/{projectname}.git" with infinite subgroups;
             var path = repositoryUri.AbsolutePath;
             var pathParts = path.Split('/')
                 .Where(s => !string.IsNullOrWhiteSpace(s))
                 .ToList();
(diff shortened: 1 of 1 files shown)


MarcBruins:
Can you add some test to this part?

Calytic:
As it is, this PR does not work.  I did test it in production on GitLab.   For example..

Say the repo is `https://gitlab.com/myrepo/mysubgroup/myrepo.git`

This PR as it is will actually have a different, but equally broken result.  The repo will validate (as opposed to the exception before) but the RepositorySetting will expect the repo to be here:

`https://gitlab.com/myrepo/myrepo.git`

So, the next change I made was to grab all the path parts after owner, so the `repoName` becomes

`mysubgroup/myrepo`

and then it works, RepositorySettings correctly finds the repo at..

`https://gitlab.com/myrepo/mysubgroup/myrepo.git`

However, `mysubgroup/myrepo` is technically _not_ the repository name.  The repository name is `myrepo` and the repository group is `mysubgroup`.  I will go ahead and commit the change that makes this work in production but in order to remain technically accurate, more substantial changes will need to be made so RepositorySettings can accommodate the _concept_ of having a repository group, which it currently does not.  Someone more familiar with the internals of NuKeeper and the implications of making such a change across all of the different drivers that NuKeeper supports should probably address this.

For this reason, and without a unit test, I don't expect this PR to be accepted even if it is a proof of concept.

PR #921, opened 2020-02-14
Fix UriFormatException

Skotina:
Fix UriFormatException with message "Invalid URI: Invalid port specified." when Username contains '@'

When Username contains '@'  throws UriFormatException  with message "Invalid URI: Invalid port specified."

UriFormatException 

Replace '@' by "%40"

No

- [x] All projects build

### DIFF
--- a/NuKeeper.Git/GitCmdDriver.cs
+++ b/NuKeeper.Git/GitCmdDriver.cs
@@ -100,7 +100,7 @@ namespace NuKeeper.Git
                 return pullEndpoint;
             }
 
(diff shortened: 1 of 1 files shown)


AnthonySteele:
Fails tests when the string is null:
```
2020-02-14T13:27:00.6094324Z    OneTimeSetUp:   Expected: No Exception to be thrown
2020-02-14T13:27:00.6095250Z   But was:  <System.ArgumentNullException: Value cannot be null. (Parameter 'stringToEscape')
```

Issue #923, opened 2020-02-19
LibGit2SharpDiscoveryDriver - File-urls are handled badly


sharpSteff:
## 🐛 Bug Report

I run the repo command against a local repository to create a PR for Gitea(this feature is currently in making)

The path is `D:\#Dev\Git\MyImportantGitRepo`

LibGit2SharpDiscoveryDriver does not recognize the folder as a git repository because of the '#'

The '#' is not Unescaped before executing  
`
 Repository.Discover(repositoryUri.AbsolutePath);
`
in `LibGit2SharpDiscoveryDriver`

### Expected behavior
LibGit2SharpDiscoveryDriver should discover the repo even if there is a special char.

replacing it with 
`Repository.Discover(Uri.UnescapeDataString(repositoryUri.AbsolutePath));`

resolved the issue
### Reproduction steps

launch nukeeper repo "LocalPathWithSpecialChar"

### Configuration
not needed

**Version:** 0.26.1

sharpSteff:
I can provide a PR if you like

PR #924, opened 2020-02-19
Issue 923

sharpSteff:
Bugfix for #923 

local paths with special characters throws an exception

local paths with special characters are working now

Should not

Running local folder check for azure/vtst

- [ ] All projects build


MarcBruins:
There seems to something missing for .NET core 3.1 and URI which seems weird to me. Needs to investigate what is going on

Issue #942, opened 2020-04-25
Update Nuget Package Failed because URL to an interal Nuget Feed contains a Space


may82:
## 🐛 Bug Report

We are happy that we found this tool. Our system configuration is special, so we encountered a rare error.

Our internal Nuget-Feed Link (e.g. http:\\tfs.local\Our%20Solution\_packaging/our-Feed/nuget/v3/index.json) contains a space. 

In line 36 (DotNetUpdatePackageCommand.cs) NuKeeper cut the Url in two parts (e.g. http:\\tfs.local\Our Solution\_packaging/our-Feed/nuget/v3/index.json) while executing "dotnet add" command. 

here is the part of the log (--verbosity D) before the exception is thrown: 
`Nuget restore on local.project.TestConsole.csproj complete
In path C:\tfs\OurSolution\local.project\Development\local.project.TestConsole, running command: dotnet restore local.project.TestConsole.csproj -s http://tfs.local/Our%20Solution/_packaging/our-Feed/nuget/v3/index.json
In path C:\tfs\OurSolution\local.project\Development\local.project.TestConsole, running command: dotnet remove local.project.TestConsole.csproj package local.package
In path C:\tfs\OurSolution\local.project\Development\local.project.TestConsole, running command: dotnet add local.project.TestConsole.csproj package local.package -v 1.0.4.209 -s http://tfs.local/Our Solution/_packaging/our-Feed/nuget/v3/index.json`

### Expected behavior

To add in line 36 masked quotes for parameter "-s" 

e.g. dotnet add local.project.TestConsole.csproj package local.package -v 1.0.4.209 -s **"**http://tfs.local/Our Solution/_packaging/our-Feed/nuget/v3/index.json**"**

AnthonySteele:
What's interesting is that the `dotnet restore` part escapes the url. I assume that the `%20` escaping there is  correct and can be applied to the `dotnet add` part as well?

AnthonySteele:
Should be fixed in latest release, can you verify?

PR #943, opened 2020-04-29
Quote package sources that contain spaces

AnthonySteele:

Fix for space in url, issue For https://github.com/NuKeeperDotNet/NuKeeper/issues/942

`DotNetUpdatePackageCommand` fails at the "dotnet add" step when the package source is a url containing a space, as this is not correctly handled by the commandline unless it is escaped.

Should not fail any more, the url should be passed to `dotnet add`
We could also escape the url with turning spaces to `%20` via `Uri.EscapeUriString`
Some of these uris will be local folder paths, not `http://` uris, so those also need quoting.
So I think that `" "` quoting is preferred - the issue is really with a command-line param containing a space, so quoting it with `" "` seems a more accurate fix.

No, it only happens on package sources with spaces in the uri.

- [x] All projects build

### DIFF
--- a/NuKeeper.Update/Process/DotNetUpdatePackageCommand.cs
+++ b/NuKeeper.Update/Process/DotNetUpdatePackageCommand.cs
@@ -49,7 +49,7 @@ namespace NuKeeper.Update.Process
 
             if (result.IndexOf(" ", StringComparison.OrdinalIgnoreCase) >= 0)
             {
-                return Uri.EscapeUriString(result);
+                return "\"" + result + "\"";
             }
 
             return result;
--- a/NuKeeper.Update/Process/DotNetUpdatePackageCommand.cs
+++ b/NuKeeper.Update/Process/DotNetUpdatePackageCommand.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Threading.Tasks;
 using NuGet.Configuration;
 using NuGet.Versioning;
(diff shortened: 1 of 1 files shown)


skolima:
In other places, code is using `ArgumentEscaper.EscapeAndConcatenate` which encapsualates it - would it work here? Example: https://github.com/NuKeeperDotNet/NuKeeper/blob/master/NuKeeper.Abstractions/NuGet/NuGetSources.cs#L57

AnthonySteele:
I'm not sure what the logic in the linked code is for - only escape if it's  a file path? Why?

AnthonySteele:
Anyway,  it looks like `ArgumentEscaper.EscapeAndConcatenate` should work, this test is green:

```csharp
[Test]
public void WhatDoesEscapeAndConcatenateDo()
{
    var packageSource1 = new PackageSource("http://tfs.local//Our Solution_packaging/our-Feed/nuget/v3/index.json");
    var packageSource2 = new PackageSource("http://tfs.local//OurSolution_packaging/our-Feed/nuget/v3/index.json");

    var quotedSource = ArgumentEscaper.EscapeAndConcatenate(new string[] { packageSource1.Source });
    var unQuotedSource = ArgumentEscaper.EscapeAndConcatenate(new string[] { packageSource2.Source });

    Assert.That(quotedSource, Is.Not.EqualTo(packageSource1.Source));
    Assert.That(quotedSource, Is.EqualTo("\"" + packageSource1.Source+ "\""));
    Assert.That(unQuotedSource, Is.EqualTo(packageSource2.Source));
}```

AnthonySteele:
The source is here: https://github.com/natemcmaster/CommandLineUtils/blob/master/src/CommandLineUtils/Utilities/ArgumentEscaper.cs

PR #967, opened 2020-05-06
Escape all sources when they need it

AnthonySteele:

Fix potential bug with remote sources where the url contains a space
The need for escaping the source in an arg doesn't depend on if it's a local (file) or remote (http/s) uri. It depends only on if there is a space in it or not. This can be in either, although it's more likely in a file path with "C:\Program Files" etc
ArgumentEscaper will check for spaces and handle it when needed. [Code](https://github.com/natemcmaster/CommandLineUtils/blob/master/src/CommandLineUtils/Utilities/ArgumentEscaper.cs)

Follow-on from https://github.com/NuKeeperDotNet/NuKeeper/pull/943
particularly https://github.com/NuKeeperDotNet/NuKeeper/pull/943#issuecomment-621322856

- [ ] All projects build

### DIFF
--- a/NuKeeper.Abstractions/NuGet/NuGetSources.cs
+++ b/NuKeeper.Abstractions/NuGet/NuGetSources.cs
@@ -45,7 +45,7 @@ namespace NuKeeper.Abstractions.NuGet
         public string CommandLine(string prefix)
         {
             return Items
-                .Select(s => $"{prefix} {EscapePathIfLocal(s)}")
+                .Select(s => $"{prefix} {EscapeSource(s)}")
                 .JoinWithSeparator(" ");
         }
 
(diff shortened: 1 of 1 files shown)


skolima:
I do remember there *was* a reason to escape only local urls. I really can't remember it and I have no test for it.

AnthonySteele:
We had to escape http urls in #943, in a different place in the code.
We can park this for now,  and see if they verify that #943 is in fact fixed now, and in case the reason come back to mind.

skolima:
I'm thinking this should be safe to merge. If there's a case which fails, we'll get a test. Maybe I'm misremembering.

AnthonySteele:
I think that the argument was that local `file://` paths can contain a space ( the "C:\Program Files" case) 
and that http urls do not.
It turns out that 1) that's not always true about http urls, and 2) `ArgumentEscaper` does the right thing at a lower level by checking for spaces

Issue #896, opened 2019-11-03
Skipping package 'Microsoft.AspNetCore.App' with no version specified in file '....csproj'.


timmkrause:
## 🐛 Bug Report

Having the following variant of a `<PackageReference>` where the version is mentioned in a sub-element:

```
    <PackageReference Include="Roslynator.Analyzers">
      <Version>2.2.0</Version>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
```
NuKeeper outputs

`Skipping package 'Roslynator.Analyzers' with no version specified in file 'D:\a\1\s\Directory.Build.props'.`

### Expected behavior

Sub-elements of a package reference containing the version should be taken into account as well.

### Configuration

**Version:** 0.25.0

**Platform if applicable:** 
- [x] :hammer_and_wrench: NuKeeper CLI
- [ ] :sparkles: GitHub
- [x] :robot: AzureDevops
- [ ] :checkered_flag: Bitbucket
- [ ] :earth_americas: Gitlab
- [ ] :tv: Gitea
- [ ] :whale: Docker

timmkrause:
Still having this issue, just seen it today with NuKeeper version 0.27.1:

`Directory.Build.props`:

```
<Project>
  <!-- Condition in place to circumvent NuGet restore error, see https://github.com/NuGet/Home/issues/8004. -->
  <ItemGroup Condition="'$(MSBuildProjectExtension)' != '.deployproj'">
    <PackageReference Include="Roslynator.Analyzers">
      <Version>2.3.0</Version>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
    <PackageReference Include="StyleCop.Analyzers">
      <Version>1.1.118</Version>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers">
      <Version>2.9.8</Version>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
  </ItemGroup>
</Project>
```

[...]

PR #971, opened 2020-05-13
Handle single-version-wide exact version range.

skolima:


AnthonySteele:
This fixes a part of https://github.com/NuKeeperDotNet/NuKeeper/issues/896

Issue #973, opened 2020-05-15
Enable BitBucket Server integration to lookup current user using API


KeyDecoder:
## 🏗 Enhancement Proposal

Currently when using BitBucket Server integration, it takes the username from the repository (ie. nukeeper.exe repo --platform bitbucketlocal https://myusername@server/scm/...).  

However when it creates a user to do the git commits, it (in GetCurrentUser() in BitBucketLocalPlatform) it only sets the username and leaves the Name and Email blank.  This means it always falls back to using the name and email in the git config.

This uses the git config if the global git config is set, but if the global git config user and email is not set, it will fail with "Failed to build signature, did not get valid git user identity from token or from repo config" error.

This has been causing us problems when running on a build agent where we don't want to set the git settings globally and be able to set the user based on the username passed in.  In this case, we are using this with a specific user for automatic updating so it's clear in the commits which user this was, but we don't want to set this in the global git settings.

### Pitch

The change is to lookup the user in BitBucket Server API so it can return the complete user name and email.   This makes it similar to GitHub and GitLab integrations.

This would use the BitBucket Server users API to lookup the user to get the name and email (https://docs.atlassian.com/bitbucket-server/rest/5.16.0/bitbucket-rest.html#idm8285279136).

### Platforms affected (mark all that apply)
- [ ] :hammer_and_wrench: NuKeeper CLI
- [ ] :sparkles: GitHub
- [ ] :robot: AzureDevops
- [x] :checkered_flag: Bitbucket (BitBucket Server only)
- [ ] :earth_americas: Gitlab
- [ ] :tv: Gitea
- [ ] :whale: Docker

PR #974, opened 2020-05-15
[Bitbucket Local] Add new API call to get the current BitBucket Server user details

KeyDecoder:
This adds a new API call for BitBucket Server (local) to get the user details from the username provided to use in creating a valid git user for the platform if possible.

Changed GetCurrentUser to use the actual user details if they can be read out instead of username and empty email which was causing it to always revert to using the git config

Enhancement to use the user details from the user provided for BitBucket Server integration.

Currently only provides the username when creating a user for a commit signature, but this fails (since no name or email) and falls back to using the git config.

It will now attempt to lookup the user in BitBucket to get the name and email. If it cannot find the user, it will still fallback to the existing behaviour.

This will break any integrations which rely on it using the git config for the user that creates the commits and if this user is different to the username provided in the command line.

No impact on any other integrations or BitBucket Cloud.

- [x] All projects build

### DIFF
--- a/NuKeeper.Git.Tests/TestDirectoryHelper.cs
+++ b/NuKeeper.Git.Tests/TestDirectoryHelper.cs
@@ -64,6 +64,16 @@ namespace NuKeeper.Git.Tests
 
         private static void NormalizeAttributes(DirectoryInfo toNormalize)
         {
+            // Workaround for https://github.com/libgit2/libgit2sharp/issues/1726
+            // where libgit2sharp creates symlink directories
+            // Note, if symlinks = false in your gitconfig then this won't happen
+            var fileAttributes = File.GetAttributes(toNormalize.FullName);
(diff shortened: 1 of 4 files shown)


KeyDecoder:
Implements #973

KeyDecoder:
Hi @AnthonySteele and other maintainers, what is the best way to progress this?  I'm not sure if have any BitBucket Local specific developers who should review this.

KeyDecoder:
Add a check to resolve an issue where libgit2 was creating symbolic links and it was causing issues in test cleanup where tests did a checkout and tried to clean up.

skolima:
For the libgit2 issue, after confirming with @AnthonySteele , I've upgraded to the preview version which doesn't have the bug. So we don't need the workaround in this ticket.

As for BitBucket maintenance/features, @MarcBruins could you perhaps take a look please?

PR #1011, opened 2020-08-22
feat(AD/TFS): Support reviewers in PRs

CrispyDrone:

Feature

None

You can now specify reviewers to be added to the pull requests created
for the Azure Devops (Server), and TFS or VSTS platforms.

You can do this by either specifying the `--reviewer` option on the
commandline, or by adding a `Reviewers` property to the
`nukeeper.settings.json` file.

As usual, CLI arguments take precedence over settings from the file
system.

The reviewer value should be an email address.

The current implementation uses the REST apis for teams, and team
members. However, I noticed there's also a users api that allows
querying based on the email address. I did not however see an easy way
to make sure that a particular user is part of a team by using that
API, as the `membersOf` property was empty in my case.

Somebody who is more familiar with these apis could probably improve
this implementation significantly but it is working for my azure devops
server instance.

No. When creating the pull request, reviewers are immediately added, however if for some reason reviewers cannot be resolved, the pull request is still created.

Azure Devops vs Azure Devops Server vs TFS vs VSTS. I only have access to a Azure Devops Server instance. I can confirm it's working there, but I haven't been able to test other cases.

?

- [X] All projects build

### DIFF
--- a/NuKeeper.AzureDevOps/AzureDevopsPlatform.cs
+++ b/NuKeeper.AzureDevOps/AzureDevopsPlatform.cs
@@ -23,14 +23,19 @@ namespace NuKeeper.AzureDevOps
             _clientFactory = clientFactory;
         }
 
-        public void Initialise(AuthSettings settings)
+        protected virtual AzureDevOpsRestClient GetClient(AuthSettings settings)
         {
             if (settings == null)
             {
                 throw new ArgumentNullException(nameof(settings));
             }
 
-            _client = new AzureDevOpsRestClient(_clientFactory, _logger, settings.Token, settings.ApiBase);
+            return new AzureDevOpsRestClient(_clientFactory, _logger, settings.Token, settings.ApiBase);
+        }
+
+        public void Initialise(AuthSettings settings)
+        {
+            _client = GetClient(settings);
         }
 
         public async Task<User> GetCurrentUser()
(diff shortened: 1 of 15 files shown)


CrispyDrone:
Rebased to accommodate new changes.

NuKeeper.AzureDevOps/AzureDevopsPlatform.cs
```diff
@@ -88,6 +94,48 @@ public async Task OpenPullRequest(ForkData target, PullRequestRequest request, I
...
             };
 
+            // todo: maybe they don't want to create reviewers during the pull request itself, but add individual reviewers later on?
+            // could be ok, if you want to make sure the pull request creation doesn't fail, but you can simply continue without reviewers
+            // anyway. I don't see any advantage to adding reviewers/labels separately.
+
+
+            // It seems there's an api to search for a user based on email address, maybe we should just use that?
+            // but I guess in theory the user should be part of the team...?
+            if (request.Reviewers.Any())
```
MarcBruins, line 104: Think we can remove this comments and documented the strategy that was chosen and why

NuKeeper.AzureDevOps/AzureDevopsRestTypes.cs
```diff
@@ -28,6 +37,47 @@ public class Creator
...
+                            return mailString;
+                    }
+                }
+
+                return string.Empty;
+            }
+        }
+    }
+
+    //public class PropertiesCollection
```
MarcBruins, line 69: Remove this if not needed

CrispyDrone:
I'll try to tackle the comments and the merge issues this week.

NuKeeper.AzureDevOps/AzureDevopsPlatform.cs
CrispyDrone: Done.

NuKeeper.AzureDevOps/AzureDevopsRestTypes.cs
CrispyDrone: Done.

WhiteBahamut:
Is it possible to add an Azure DevOps group/team instead of email? It would be great to be able to set a team as the reviewer, like in the branch policies "automatically include code reviewers" section

CrispyDrone:
Hi @WhiteBahamut 

I will have a look if I can update the implementation. I encountered some issues after having migrated to Azure Devops Services from Azure Devops Server that would have to be ported from my fork anyway.

Although, to be honest, I think the code base is making it quite difficult to be productive. I think a significant refactoring needs to happen and better sooner than later. The reason why I mention this is because in my own fork I just went with a very dumb hack to solve the problem of Azure Devops Services vs Server because it is getting too complex to code a proper solution.

CrispyDrone:
Azure Devops Services uses a different url i.e. `https://vssps.dev.azure.com` as domain instead of the expected `https://dev.azure.com`.

WhiteBahamut:
@CrispyDrone thanks for the update. Found a solution for my org. We setup a branch policy to include default reviewers in PRs on branches. So when NuKeeper triggers a PR, the group is added by Azure anyways. PR is displayed for the team to review/approve.

Issue #1004, opened 2020-07-18
NuKeeper hangs on NuGet restore command


CrispyDrone:
I just started playing with NuKeeper again today since I noticed some new activity on the project.

I upgraded to the new version, and tried to rerun the same commands I did after correcting for some now non-existent parameters etc.

I'm currently stuck with the following issues:
1. my `.nuget/NuGet.Config` file isn't getting picked up any more by NuKeeper (see https://github.com/NuKeeperDotNet/NuKeeper/issues/907)
2. I'm either receiving `Unable to load the service index for source` or `401 unauthorized` for packages of my private nuget feed (see https://github.com/NuKeeperDotNet/NuKeeper/issues/542).
3. the nuget restore command hangs indefinitely

Point 1 I can solve by changing `%APPDATA%\NuGet\NuGet.Config` to include our private nuget feed or passing in the feeds manually using `--source`.

Point 2 I have exhausted all options. I've experimented with `http_proxy` settings and `packageSourceCredentials` using all possible combinations of my PAT  (with all scopes enabled), plaintext password, my username, my domainname,... nothing works.

I'm not running behind a proxy as far as I know, I'm connected to my corporate network through a VPN however.

We recently upgraded to Azure Devops Server 2019 from an older TFS2017, and just this morning I had some issues with cloning repositories that was solved by __not__ using the `https` version of the url see https://github.com/microsoft/Git-Credential-Manager-for-Windows/issues/904

When I run the nuget restore command—that NuKeeper tries to invoke—manually, and remove the `-NonInteractive` flag, it will prompt me for a username and password, and subsequently successfully restore all packages.

_Originally posted by @CrispyDrone in https://github.com/NuKeeperDotNet/NuKeeper/issues/542#issuecomment-660492017_

CrispyDrone:
I retried it today using visual studio's devenv command line and it seems to work! See https://github.com/NuKeeperDotNet/NuKeeper/issues/916 In this however I was actually using windows cmd and not bash that's included with a git installation on windows.

There are still errors related to `Cannot determine the packages folder to restore NuGet packages. Please specify either -PackagesDirectory or -SolutionDirectory.`. I don't know the significance of this however since pull requests are being created.

CrispyDrone:
Bump.

CrispyDrone:
Fixed by #1044

PR #1044, opened 2020-10-04
Prevent external processes from hanging forever

CrispyDrone:

Bug fix.

In certain cases, invoking an external process will result in NuKeeper hanging forever waiting for it to terminate. This can reliably be reproduced (in my case), when debugging NuKeeper. When invoking NuKeeper from the command line (without a debugger attached?), this does not happen.

No.

Try debugging NuKeeper. For me this consistently hangs on the first restore solution command. This also happens when running it from bash. It does not happen when I run from the windows cmd console (perhaps after invoking VsDevCmd.bat, since I've been doing that to prevent other problems).

+ #1004 

No relevant documentation, I think.

- [X] All projects build

### DIFF
--- a/NuKeeper.Update/ProcessRunner/ExternalProcess.cs
+++ b/NuKeeper.Update/ProcessRunner/ExternalProcess.cs
@@ -1,8 +1,8 @@
+using NuKeeper.Abstractions;
+using NuKeeper.Abstractions.Logging;
 using System;
 using System.Diagnostics;
 using System.Threading.Tasks;
-using NuKeeper.Abstractions;
-using NuKeeper.Abstractions.Logging;
 
 namespace NuKeeper.Update.ProcessRunner
 {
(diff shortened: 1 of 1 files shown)


scp-mb:
Ran into this issue today when we had some unit test projects that failed to restore packages, so stdout was written to first, which would cause our nukeeper CI pipeline to hang until the 1 hour timeout.

skolima:
I've intermittently seen the error myself (though never managed to reproduce), but I'm confused how this change would fix it. `Task.WhenAll` requires tasks to finish without regard to the ordering, but still requires both of them to finish. How does this resolve the issue of hanged behaviour?

And could you please include a test for this change?

Greybird:
> 
> 
> I've intermittently seen the error myself (though never managed to reproduce), but I'm confused how this change would fix it. `Task.WhenAll` requires tasks to finish without regard to the ordering, but still requires both of them to finish. How does this resolve the issue of hanged behaviour?
> 
I think the key thing is that they can finish in ANY order. The previous code is waiting to read all the output stream, then waits for the error stream to be completely read.
These streams have finite size, and the child executable could have filled the error stream, waiting for it be have some room to write more. 
However, as the parent process will be waiting to have finished reading the output stream (which is not the case as the child process could write to it afterwards), before reading the error stream, both process are deadlocked.

You can find some details about these caveats of these streams in the [ms doc](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standarderror?view=net-5.0), when in the last example, the error stream is read using the asynchronous method (ie events) while the output is read synchronoulsy (these synchronous/asynchronous not to be confused with async/await usage)

skolima:
This (and the MSDN docs) make sense now. Thanks.

PR #1057, opened 2020-11-08
fix(logging): Prevent IO ex concurrent writes

CrispyDrone:

Bug fix. 

`File.AppendText` takes an exclusive write lock on a file. This means
that when multiple writes are occurring at the same time, an
`IOException` is thrown saying: "The process cannot access the file X
because it is being used by another process".

No. 

This fix will only work as long as the logger is used as a "singleton" throughout the application as the lock is not a static property. I decided against making it static because the constructor has file path as a parameter.

Use `--logfile`. This caused issues especially in the package update lookup process as it's parallelized.


skolima (approved):

skolima:
Looks good. Thank you for your contribution.

PR #1058, opened 2020-11-08
Nukeeper/fix/return non zero code when updates fail

CrispyDrone:

Bug fix.

In many locations exceptions were being swallowed inappropriately. 

I've ensured that each separate repository gets a chance to run in case
of a multiple repository update command; single repository commands on
the other hand will now fail as soon as one of its update sets cannot be
applied.

The reasoning for this is the following: an error at the level of an
update set being applied is most likely going to happen for all the
different update sets. It is after all working with the same git
repository, target repository,... For example, if a push fails due to an
authentication issue, it will happen for all update sets.

After all repositories have had their chance and at least one of them
fails, an exception is throw that will result in a non-zero status code
being returned to process that invoked nukeeper.

`ICollaborationEngine` returns an integer, but this represents the
number of updates it has performed, and since there's no easy way to know
what the expected number of updates is at this point, there's no way to
utilize this information to be able to tell whether
`CollaborationPlatformCommand` should return a non-zero status code.

Yes. NuKeeper will now return a non-zero exit code in case of failure.

Compare behavior of various commands before and after to ensure NuKeeper is return a non-zero status code at the appropriate times.

### DIFF
--- a/NuKeeper.Tests/Commands/RepositoryCommandTests.cs
+++ b/NuKeeper.Tests/Commands/RepositoryCommandTests.cs
@@ -1,10 +1,4 @@
-using System;
-using System.Collections.Generic;
-using System.Reflection;
-using System.Threading.Tasks;
-
 using NSubstitute;
-
 using NuKeeper.Abstractions.CollaborationModels;
 using NuKeeper.Abstractions.CollaborationPlatform;
 using NuKeeper.Abstractions.Configuration;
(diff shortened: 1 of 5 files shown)


AnthonySteele (approved):

skolima:
@CrispyDrone could you please include a test for this change?

CrispyDrone:
@skolima 

I've added a few tests for `CollaborationEngine`. I also noticed I introduced an error in terms of the folder cleanup, which I've fixed through adding a try/finally statement.

If you can think of any more scenarios you would like tested, feel free to share.

PR #1076, opened 2020-12-08
Update NuKeeper.csproj to enable .NET 5.0 environments to use the tool

rajbos:
Update to support .NET 5.0 based environments

Adds support in the NuKeeper project for usage of the dotnet global tool in a pure .NET 5.0 environment.

Running in a docker container with only .NET 5.0 installed, this is the current output:

```
Tool 'nukeeper' (version '0.34.0') was successfully installed.
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '3.1.0' was not found.
  - The following frameworks were found:
      5.0.0 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]
```
So we cannot use NuKeeper without also installing .NET Core 3.1

With this new target framework version `dotnet tool install` will recognize the correct TFM to use and fix the issue.

I don't think so :-)

- [x] All projects build

### DIFF
--- a/NuKeeper/NuKeeper.csproj
+++ b/NuKeeper/NuKeeper.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1;net50</TargetFrameworks>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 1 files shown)


AnthonySteele (approved):

rajbos:
Unfortunately you'll need to update the build pipelines as well to use a new base image. Would love to help if needed 👍

MarcBruins:
@rajbos Think #1084 also consumes does the same as this PR, i would prefer merging that one

rajbos:
Great, hope to see the .NET 5.0 support soon :-)

PR #1085, opened 2021-01-11
Added support for GitPullRequestMergeStrategy for Azure DevOps

jelledejong:
Added support for GitPullRequestMergeStrategy for Azure DevOps

By default the merge behaviour is noFastForward, if this is disabled by policies, then it will not auto-merge

Repo command expanded with option MergeStrategy, with the following options:
- noFastForward,
- rebase,
- rebaseMerge,
- squash

no, it's been made an optional parameter with default noFastForward, which was its original default

- [ ] All projects build

### DIFF
--- a/NuKeeper.Abstractions/Configuration/GitPullRequestMergeStrategy.cs
+++ b/NuKeeper.Abstractions/Configuration/GitPullRequestMergeStrategy.cs
@@ -2,9 +2,9 @@ namespace NuKeeper.Abstractions.Configuration
 {
     public enum GitPullRequestMergeStrategy
     {
-        noFastForward,
-        rebase,
-        rebaseMerge,
-        squash
+        NoFastForward = 0,
+        Rebase = 1,
+        RebaseMerge = 2,
+        Squash = 3
     }
 }
--- a/NuKeeper.Abstractions/Configuration/GitPullRequestMergeStrategy.cs
+++ b/NuKeeper.Abstractions/Configuration/GitPullRequestMergeStrategy.cs
@@ -1,0 +1,10 @@
+namespace NuKeeper.Abstractions.Configuration
+{
+    public enum GitPullRequestMergeStrategy
+    {
+        noFastForward,
+        rebase,
+        rebaseMerge,
+        squash
+    }
+}
--- a/NuKeeper.Gitea/GiteaSettingsReader.cs
+++ b/NuKeeper.Gitea/GiteaSettingsReader.cs
@@ -68,7 +68,7 @@ namespace NuKeeper.Gitea
             settings.Token = Concat.FirstValue(envToken, settings.Token);
         }
 
-        public async Task<RepositorySettings> RepositorySettings(Uri repositoryUri, bool setAutoMerge, string targetBranch = null, GitPullRequestMergeStrategy gitPullRequestMergeStrategy = GitPullRequestMergeStrategy.noFastForward)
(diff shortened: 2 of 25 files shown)


MarcBruins:
Thanks so far! I'll do a review tommorow, can you also update the docs?

NuKeeper.Abstractions/Configuration/GitPullRequestMergeStrategy.cs
```diff
@@ -0,0 +1,10 @@
+namespace NuKeeper.Abstractions.Configuration
+{
+    public enum GitPullRequestMergeStrategy
+    {
+        noFastForward,
```
MarcBruins, line 5: Please use PascalCase and make sure that the default has a value of 0

NuKeeper.Gitea/GiteaSettingsReader.cs
```diff
@@ -68,7 +68,7 @@ public void UpdateCollaborationPlatformSettings(CollaborationPlatformSettings se
             settings.Token = Concat.FirstValue(envToken, settings.Token);
         }
 
-        public async Task<RepositorySettings> RepositorySettings(Uri repositoryUri, bool setAutoMerge, string targetBranch = null)
+        public async Task<RepositorySettings> RepositorySettings(Uri repositoryUri, bool setAutoMerge, string targetBranch = null, GitPullRequestMergeStrategy gitPullRequestMergeStrategy = GitPullRequestMergeStrategy.noFastForward)
```
MarcBruins, line 71: Do other platforms have these mergeoptions?

jelledejong: We hadn't looked into that, we noted in the docs that we made it work for Azure only. We copied the implementation of setAutoMerge which is also similar (basically a param that gets pushed down and is Azure specific)

NuKeeper.Abstractions/Configuration/GitPullRequestMergeStrategy.cs
jelledejong: Will do

NuKeeper.Gitea/GiteaSettingsReader.cs
MarcBruins: So now we have azure specific merging in all platforms? That doesn't seem right

jelledejong: I agree with the sentiment that it doesn't feel right, I too felt a bit unclean after doing this. Yet at the same time it is platform-specific because it is tied to automerge which is also already platform specific (a merge-strategy doesn't make much sense if it isn't an auto merge because a manual merge chooses the way they merge). So in the end we decided to follow the implementation of the automerge because they are that closely tied together.

I also took a look at passing down the variable another way, rather than populating the generic RepositorySettings.cs, we looked at simply passing the variable down to where it is actually needed and use it there, that would mean passing it down to the PackageUpdater which is also quite generic and doesn't quite make sense. So in the end we decided to keep the implementation close to the implementation of the automerge as that seemed more preferable.

I guess it would also be an option to refactor the code and push platform-specific data to the platforms, but that still requires casting down the line somewhere. Generally I don't favor that approach.

It might also be an option to change the enum itself to a string and leave the field open to the user. Don't validate it at all, and just pass it along to the API hoping the consumer put in a correct string. This does have the upside of having a more flexible tool (if azure changes the api, then you can pass updated variables without updating nukeeper). It doesn't really solve the platform-specific option except that it looks less platform specific.

Finally I guess we can create a new azure devops command specifically, but that obviously requires a little bit more thought than our issue at hand: we want to set the merge-strategy for azure devops because we otherwise can't make use of the automerge option ;)

Either way, we are dealing with a platform specific parameter because of the 'reliance on automerge which is already AzureDevOps specific'-argument.

PR #1108, opened 2021-05-12
Add relevant reviewers to PRs to local BitBucket repositories instead of everyone in the Project

arikalish:
Bug fix? I guess?

When creating a PR to a locally-hosted BitBucket repository everyone set as a default reviewer in the Project, including users that no longer exist, are added as reviewers.

This results in people who don't need to see the PRs being added to them erroneously.

Also, default reviewer rules are not respected, so folks that should be put on PRs are not added.

If a user no longer exists NuKeeper fails due to the presence of an error section in the response, even though BitBucket also provides a list of users.

For reference, repositories live inside of projects in BB, so we might have:

Project X
  - Repo

If Project X has reviewers:
A: All branches
B: Branches named "foo-*"
C: Branches named "bar-*"
Repo:
D: "foo-*"

Example cases:

PR of branch "abc" into Repo:
Users A, B, and C will be added to the PR but user D will not be.
Expected behavior in that case would be for user A to be added and no one else.

PR of branch "foo-123" into Repo:
Users A, B, and C will be added to the PR but user D will not be.
Expected behavior: Users A, B, and D are added.

PR of branch "bar-456" into Repo:
Users A, B, and C will be added to the PR but user D will not be.
Expected behavior: Users A and B are added.

Call the default reviewers API on the BitBucket instance to get the list of expected reviewers from BitBucket itself and filter out any inactive (aka: deleted) users that might have been left in rules on the server. BitBucket will apply the rules that would be applied if a user had created the PR manually.

Yes, but the existing behavior was likely not what was intended

See above example cases, also test with users that have been removed from BitBucket but are still present in default reviewer rules.

N/A

- [X] All projects build

### DIFF
--- a/Nukeeper.BitBucketLocal/BitBucketLocalPlatform.cs
+++ b/Nukeeper.BitBucketLocal/BitBucketLocalPlatform.cs
@@ -66,7 +66,7 @@ namespace NuKeeper.BitBucketLocal
             var repositories = await _client.GetGitRepositories(target.Owner);
             var targetRepository = repositories.FirstOrDefault(x => x.Name.Equals(target.Name, StringComparison.InvariantCultureIgnoreCase));
 
-            var reviewers = await _client.GetBitBucketReviewers(target.Owner, targetRepository.Name);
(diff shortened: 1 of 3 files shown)


arikalish:
Still relevant

skolima:
Looks good. Thanks.

PR #1127, opened 2021-09-20
Drop netcoreapp2.1 target, it's no longer supported

skolima:
Adds a .NET SDK 5.0 target to the Docker images published (after this is merged, I'll update Docker Hub configuration).


msallin (approved):
Lgtm!

PR #1138, opened 2021-09-20
Simplify NuGet CLI reference

slang25:


msallin (approved):

skolima:
It works! Oh man that is so much nicer than having to update it by hand every time.

PR #1141, opened 2021-09-22
Bump version in Dockerfile to 0.35.0

skolima:
Update to current release in the Dockerfiles provided.

### DIFF
--- a/Docker/SDK3.1/Dockerfile
+++ b/Docker/SDK3.1/Dockerfile
@@ -1,5 +1,5 @@
 FROM mcr.microsoft.com/dotnet/core/sdk:3.1-bionic
-ARG NUKEEPER_VERSION=0.34.0
+ARG NUKEEPER_VERSION=0.35.0
 RUN dotnet tool install --global NuKeeper --version $NUKEEPER_VERSION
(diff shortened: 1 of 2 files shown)


msallin:
Is this necessary, can't we just use the latest?

skolima:
Problem with having `latest` in there is that the Docker images aren't deterministic then.

And building them becomes problematic, as it's either "invalidate every time and build from scratch" or "we'll get whatever is cached on the build machine, perhaps the current version, perhaps 3 versions stale".

msallin:
I see. How do we release a new NuKeeper version? Maybe we can automate this.
I'm working on getting a bit of dev power from our organization, to do such things.

msallin (approved):

skolima:
There's a (somewhat grotty) Azure Devops pipeline. It's not the most readable, TBH. I'd prefer GitHub actions, or really anything that's versioned in the repository as a text file. The Docker images used to be released automatically by Docker Hub - but now they've changed their licensing, there's no longer a free build tier for Open Source - so the last ones was built on my machine. That could also move to GitHub Actions - there's free automation here for it.

